PR #62, opened 2016-09-06
Fixing a panic in Ec2Info.go

marcboudreau:
Initializing the metaClient field with NewEc2Meta() instead of with a struct literal, so that the cache field in the embedded Ec2Meta instance is properly initialized.

I looked at why the tests didn't catch this and it appears that this would be a difficult case to catch in a non-EC2 environment.  Specifically, https://github.com/hairyhenderson/gomplate/blob/master/aws/ec2info.go#L62 short circuits the code path that would have caused the test to fail.

### DIFF
--- a/aws/ec2info.go
+++ b/aws/ec2info.go
@@ -20,7 +20,7 @@ type InstanceDescriber interface {
 
 // NewEc2Info -
 func NewEc2Info() *Ec2Info {
-	metaClient := &Ec2Meta{}
+	metaClient := NewEc2Meta()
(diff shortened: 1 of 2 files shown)


hairyhenderson:
@marcboudreau awesome, thanks for the contribution! Can you add a test that covers the bug?

marcboudreau:
I've added a test to ec2info_test.go to cover this case.

hairyhenderson:
LGTM! Thanks again for the contribution! ⭐

PR #73, opened 2016-10-13
Added a read me section about multiple line if/else/end statements.

EtienneDufresne:
Added a read me section about multiple line if/else/end statements.


hairyhenderson:
Thanks for the PR, @EtienneDufresne!

Issue #54, opened 2016-09-01
Support Vault authentication on HTTPS datasource


hairyhenderson:
It'd be cool to use the `https` datasource to grab secrets from HashiCorp's [Vault](https://vaultproject.io).

A bunch of auth backends exist, but I'd like at least the following to work:
- [x] [Username & Password](https://www.vaultproject.io/docs/auth/userpass.html) - #104
- [x] [AppID](https://www.vaultproject.io/docs/auth/app-id.html) - #74
- [ ] [AppRole](https://www.vaultproject.io/docs/auth/approle.html) - #105
- [x] [GitHub](https://www.vaultproject.io/docs/auth/github.html) - #80
- [x] [Token](https://www.vaultproject.io/docs/auth/token.html) - #74 

The user interface could look something like:

``` console
$ gomplate --vault-app-id=abcd1234 --vault-user-id=defg4567 \
    --datasource vault=https://my.vault.server:8200/v1/secret/supersecret \
    "my secret is: {{ (datasource "vault").data.value }}"
my secret is: supersecretvalue!
```

PR #74, opened 2016-11-14
Support for Vault datasources (app-id & token auth)

hairyhenderson:
Partial implementation of #54

This adds support for a new URL scheme `vault://` to allow reading from Hashicorp's [Vault](https://vaultproject.io), to allow injection of secrets into templates rendered by gomplate.

Right now this only supports the [`app-id`](https://www.vaultproject.io/docs/auth/app-id.html) auth backend and pre-defined tokens.

The new `vault://` URL scheme represents the base path for the vault datasource, to make it simpler to scope secrets from the same base namespace. A further path can be specified as the second argument to the `datasource` function.

For example:

```console
$ export VAULT_ADDR=https://myvaultserver.example.com:8200
$ echo '{{(datasource "vault" "secret/password").value}}' | gomplate -d vault=vault:///
foo
$ echo '{{(datasource "vault").value}}' | gomplate -d vault=vault:///secret/password
```

Right now the only way to specify the vault server address is by setting the `VAULT_ADDR` environment variable.

To use the `app-id` auth backend, set the `VAULT_APP_ID` and `VAULT_USER_ID` environment variables.

To use a predetermined token, either set the `VAULT_TOKEN` environment variable, or put the token in `~/.vault-token`. The `vault auth` command can also be used to populate `~/.vault-token`.

I'm hoping to release gomplate v1.2 with this feature, but I'm looking for feedback on the implementation details!

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ Hello world
 
 Parses a given datasource (provided by the [`--datasource/-d`](#--datasource-d) argument).
 
-Currently, `file://`, `http://` and `https://` URLs are supported.
+Currently, `file://`, `http://`, `https://`, and `vault://` URLs are supported.
 
 Currently-supported formats are JSON and YAML.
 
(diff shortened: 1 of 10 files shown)


drmdrew:
Love the idea... it makes a lot of sense since most of the time we are reading secrets from vault to pass _into_ gomplate anyway.

What about a `revoke-self` to revoke the token (when it isn't being passed in) when the gomplate run is finished?

hairyhenderson:
@drmdrew good point - I was going to add that in, but I forgot about it 😉 will add that in...

hairyhenderson:
@drmdrew PTAL - it revokes tokens now 😁

Issue #75, opened 2016-11-18
Handle vault HTTP redirects


drmdrew:
Prior to vault 0.6.2, vault standby nodes would redirect HTTP to the active vault node. Current versions of the go http package do not seamlessly handle this redirection properly.

Attempting to use gomplate with pre-0.6.2 vault in HA configuration results in a panic, e.g.
```
echo -n '{{(datasource "vault" "secret/my_secret").value}}'  | gomplate -d vault="vault:///"
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
```

drmdrew:
The relevant portions of a sample panic that occurs against pre-0.6.2 vault HA setup when a redirect from a standby is encountered:

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
       	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
       	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xf0b1a]

goroutine 1 [running]:
panic(0x51b940, 0xc4200161c0)
       	/usr/local/Cellar/go/1.7.3/libexec/src/runtime/panic.go:500 +0x1a1
github.com/urfave/cli.HandleAction.func1(0xc4201e5d00)
       	/Users/dee/gocode/src/github.com/urfave/cli/app.go:478 +0x247
panic(0x51b940, 0xc4200161c0)
       	/usr/local/Cellar/go/1.7.3/libexec/src/runtime/panic.go:458 +0x243
text/template.errRecover(0xc4201e5730)
       	/usr/local/Cellar/go/1.7.3/libexec/src/text/template/exec.go:140 +0x2ad
panic(0x51b940, 0xc4200161c0)
       	/usr/local/Cellar/go/1.7.3/libexec/src/runtime/panic.go:458 +0x243
github.com/hairyhenderson/gomplate/vault.(*Client).Login(0xc4201d2180, 0x5c3b3e, 0x7fff5fbff61f)
       	/Users/dee/gocode/src/github.com/hairyhenderson/gomplate/vault/client.go:59 +0x2a
main.readVault(0xc420074d80, 0xc4201ce040, 0x1, 0x1, 0xc420077d10, 0x1, 0x1246c, 0x7fff5fbff619, 0xc4201564bf)
       	/Users/dee/gocode/src/github.com/hairyhenderson/gomplate/data.go:244 +0x19a
[...]

PR #76, opened 2016-11-18
Handle redirects from vault server versions earlier than v0.6.2

drmdrew:
Fix issue #75.

There are two golang HTTP related issues that are involved here, due to the way golang 1.7 http package reacts to HTTP redirects:
- https://github.com/golang/go/issues/4800
- https://github.com/golang/go/issues/7912 or https://github.com/golang/go/issues/9348

The first issue causes the `X-Vault-Token` header to be dropped when following redirects of HTTP GET requests. The second issue causes redirects on POST requests not to be followed _at all_.

It does look like golang 1.8 will address both these issues once it is released... so these fixes are just needed in the meantime.

TODO:
- [x] manual testing with vault HA-server configuration w/redirects
- [x] updated tests

**PTAL**: @hairyhenderson

### DIFF
--- a/vault/app-id_strategy.go
+++ b/vault/app-id_strategy.go
@@ -28,20 +28,34 @@ func NewAppIDAuthStrategy() *AppIDAuthStrategy {
 	return nil
 }
 
-// GetToken - log in to the app-id auth backend and return the client token
-func (a *AppIDAuthStrategy) GetToken(addr *url.URL) (string, error) {
+// GetHTTPClient configures the HTTP client with a timeout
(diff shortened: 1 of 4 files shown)


hairyhenderson:
@drmdrew can you run `gofmt -s -w .` on this? As well as make sure that `golint ./...` passes. Otherwise LGTM

drmdrew:
Re-based to fixup `gofmt` and `golint` issues.

filiptodoric:
You guys are the best 👍

Issue #83, opened 2016-11-20
gomplate with vault datasource panics when environment variables are unset


drmdrew:
gomplate version 1.2.2 with vault datasources will panic in `VAULT_ADDR` is unset or if other auth strategy failures occur.

Example:
```
unset VAULT_ADDR
echo -n '{{(datasource "vault" "secret/foo").value}}' | /usr/local/Cellar/gomplate/1.2.2/bin/gomplate -d vault="vault://"
```

Results in a panic:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
       	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
       	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xf0f0a]
```

hairyhenderson:
@drmdrew I don't see a panic... 

```
$ bin/gomplate -v
gomplate version 1.2.2
$ echo -n '{{(datasource "vault" "secret/foo").value}}' | bin/gomplate -d vault="vault://"
2016/11/21 09:19:38 Couldn't read datasource 'vault': Get /v1/secret/foo: unsupported protocol scheme ""
```

Now, the error message is _totally_ lacking, but that's a different issue

hairyhenderson:
(and that's with no `VAULT_*` environment variables)

```
$ env | grep VAULT | wc -l
0
```

drmdrew:
Weird... I'm still on El Capitan... let me see how gomplate 1.2.2 reacts in a `debian:jessie` container.

drmdrew:
Nope. 💥 
`docker run -it golang bash`

```
Unable to find image 'golang:latest' locally
latest: Pulling from library/golang
386a066cd84a: Already exists
75ea84187083: Pull complete
88b459c9f665: Pull complete
a31e17eb9485: Pull complete
457559cc1d69: Pull complete
47fe51a74a06: Pull complete
08dacccac43c: Pull complete
Digest: sha256:340212e9c5d062f3bfe58ff02768da70234ea734bd022a357ee6be2a6d963505
Status: Downloaded newer image for golang:latest
root@e59465791a99:/go# go get github.com/hairyhenderson/gomplate
root@e59465791a99:/go# go install github.com/hairyhenderson/gomplate
root@e59465791a99:/go# gomplate -v
gomplate version 0.0.0
[...]

PR #84, opened 2016-11-20
No envs; don't panic

drmdrew:
Fix for #83 

When gomplate is used correctly everything works fine. However, if gomplate vault datasource is misconfigured for some reason, it tends to panic rather than exit gracefully.

This PR improves the error handling to be more graceful if `VAULT_ADDR` is unset or if other auth strategy related errors occur.

**BEFORE**:
```
└> echo -n '{{(datasource "vault" "secret/foo").value}}' | /usr/local/Cellar/gomplate/1.2.2/bin/gomplate -d vault="vault://"
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
```

**AFTER**:
```
└> echo -n '{{(datasource "vault" "secret/foo").value}}' | gomplate -d vault="vault://"
2016/11/20 14:06:42 VAULT_ADDR is not a valid URL:  <nil>
```

### DIFF
--- a/vault/client.go
+++ b/vault/client.go
@@ -28,30 +28,51 @@ type AuthStrategy interface {
 	Revokable() bool
 }
 
+// GetenvFunc returns the supplied environment variable value
+type GetenvFunc func(string) string
+
 // NewClient - instantiate a new
 func NewClient() *Client {
-	u := getVaultAddr()
-	auth := getAuthStrategy()
-	return &Client{u, auth, "", nil}
-}
-
-func getVaultAddr() *url.URL {
-	vu := os.Getenv("VAULT_ADDR")
-	u, err := url.Parse(vu)
+	client, err := NewClientUseGetenv(os.Getenv)
 	if err != nil {
-		log.Fatal("VAULT_ADDR is an unparseable URL!", err)
+		log.Fatal(err)
 	}
-	return u
+	return client
 }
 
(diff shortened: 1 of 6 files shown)


vault/client.go
```diff
@@ -28,30 +28,51 @@ type AuthStrategy interface {
 	Revokable() bool
 }
 
+// GetenvFunc returns the supplied environment variable value
+type GetenvFunc func(string) string
```
drmdrew: Admittedly... injecting a `GetenvFunc` was a _long_ way to go to avoid having to monkey with the environment variables in the tests. I went down this rabbit hole largely because I have unsetting/setting env vars in unit tests... it doesn't give me warm fuzzies.

drmdrew: Maybe I just need to get over my aversion to Setenv in tests. I see os_test.go itself does it:

https://github.com/golang/go/blob/master/src/os/os_test.go#L1118

hairyhenderson:
Based on discussion in #83 and in person I think this PR should be scoped down to just error in `getVaultAddr()` when `VAULT_ADDR` isn't set. That's enough to fix the issue... A default `AuthStrategy` would also be nice, but probably best in a separate PR

drmdrew:
In light of your comment @hairyhenderson, which I agree with, it will be simpler for me to close this PR and start over with a clean slate.

PR #85, opened 2016-11-22
No vault addr dont panic

drmdrew:
A second attempt at a fix for #84. PTAL @hairyhenderson 

This time, rather than twisting gomplate into a pretzel trying to propagate errors, a small testing hack is being made to allow `log.Fatal` to be overwritten inside tests to inject a mock/spy that doesn't cause the process to exit.

Added unit tests and also checked on the command-line that gomplate doesn't panic when some/all of the `VAULT_*` environment variables are missing:

```
echo -n '{{(datasource "vault" "secret/foo").value}}' | gomplate -d vault="vault://"
2016/11/22 09:49:45 VAULT_ADDR is an unparseable URL!

export VAULT_ADDR=https://localhost:8500
echo -n '{{(datasource "vault" "secret/foo").value}}' | gomplate -d vault="vault://"
2016/11/22 09:49:37 No vault auth strategy configured
```

### DIFF
--- a/vault/client.go
+++ b/vault/client.go
@@ -12,6 +12,9 @@ import (
 	"time"
 )
 
+// logFatal is defined so log.Fatal calls can be overridden for testing
+var logFatal = log.Fatal
+
 // Client -
 type Client struct {
 	Addr *url.URL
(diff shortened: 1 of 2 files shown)


vault/log.go
```diff
@@ -0,0 +1,23 @@
+package vault
```
hairyhenderson: can you throw these into `client_test.go`?

drmdrew:
Removed `log.go`

vault/client.go
```diff
@@ -81,9 +89,13 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 
 // Login - log in to Vault with the discovered auth backend and save the token
 func (c *Client) Login() error {
+	if c.Auth == nil {
```
hairyhenderson: Can you move this check inside `getAuthStrategy()`?

drmdrew:
Moved `logFatal` to `getAuthStrategy`

PR #94, opened 2017-02-03
Adding datasourceExists function

### LINKED ISSUE #93: Crash when datasource is not specified
Currently, when a datasource is specified in the template, but not on the CLI, gomplate crashes with a nil dereference error, which can easily be reproduced with:

`echo '{{(datasource "test")}}' | ./gomplate`

Tested with gomplate version 1.2.4 on Ubuntu 14.04.

In my opinion it shouldn't crash like that in the first place, but it would also be nice if there was a possibility to test within the template if specific datasources are specified/available before trying to access them.

Right now I solved this by wrapping gomplate and assigning a dummy file/source externally - but that's not really a clean solution in my opinion.

hairyhenderson:
Fixes #93 

This does 2 things:

1. Teaches `datasource` to be a bit more friendly when an undefined datasource is referenced:
```console
$ echo '{{datasource "test"}}' | gomplate
2017/02/02 21:09:46 Undefined datasource 'test'
```
2. Adds a new `datasourceExists` function to test whether or not a given datasource was defined:
```console
$ echo '{{if (datasourceExists "test")}}{{datasource "test"}}{{else}}no worries{{end}}' | gomplate
no worries
```


hairyhenderson:
@bartmeuris what do you think of this as a fix for #93?

bartmeuris:
Looks good, pretty much what I had in mind. Sorry I couldn't submit a PR myself, currently a bit time-constrained. Code looks pretty clean and straight-forward :)

hairyhenderson:
@bartmeuris 👍

PR #107, opened 2017-03-12
Supporting *_FILE env vars for vault datasource credentials

### LINKED ISSUE #106: Make all secrets settable via files
With new Docker Swarm Secrets, and when running `gomplate` inside a container running in a Swarm service, it'd be more secure for `gomplate` to be able to read secrets from files instead of environment variables. See https://docs.docker.com/engine/swarm/secrets/#build-support-for-docker-secrets-into-your-images

I'm thinking of pretty much everything passed through an environment variable, but especially:

- `VAULT_APP_ID`
- `VAULT_USER_ID`
- `VAULT_AUTH_GITHUB_TOKEN`
- `VAULT_AUTH_USERNAME`
- `VAULT_AUTH_PASSWORD`
- `VAULT_ROLE_ID`
- `VAULT_SECRET_ID`
- `VAULT_TOKEN` - even though `~/.vault-token` is already read by default, this one is still valuable since swarm secrets can only be exposed at `/run/secrets` (because they're in a `tmpfs` in-memory filesystem)

Basically each one of these would have an equivalent `*_FILE` version that would point to a file to be read.

hairyhenderson:
Closes #106

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ endef
 clean:
 	rm -Rf $(PREFIX)/bin/*
 	rm -f $(PREFIX)/test/integration/gomplate
+	rm -f $(PREFIX)/test/integration/mirror
 
 build-x: $(patsubst %,$(PREFIX)/bin/$(PKG_NAME)_%,$(platforms))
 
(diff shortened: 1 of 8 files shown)


hairyhenderson:
I screwed up a bunch of stuff #108 so this needs to wait until I _properly_ refactor the auth strategies...

Issue #117, opened 2017-04-17
Support for bulk operations


rhuss:
Thanks a lot for this neat tool, really like the abstraction between input, datasources and output.

It would be awesome if gomplate would support _bulk operations_ directly. E.g. when an argument `--input-dir` is given, then process all files (recursively) and store it to an `--output-dir` (maintaining the directory structure). With a `--datasource-dir` one could provide an directory for the datasources, where each file in this directory is used a datasource, where the basename is taken as the name of the datasource.

A use case for this is processing a hierarchy of configuration files whithout the need of iterating over each file.

[Here](https://github.com/bibryam/k8spatterns/blob/master/docker/gomplate/gomplate_wrapper.sh) you can find an external solution which loops over the files via a shell script. would be awesome if this could be performed from within gomplate so that one could even use `scratch` as base image.

rhuss:
If you think this is a good idea, I can work on a PR.

hairyhenderson:
> Thanks a lot for this neat tool

You're very welcome! 😁

PR #114 just went in, which allows for multiple _files_ to be read at a time (and multiple outputs specified), so that's _part_ of the equation.

But iterating dirs natively is a really good idea. I'd love to see a PR for this. I like `--input-dir` and `--output-dir` as option names.

As far as the collection of datasources goes, how about instead having a new URL scheme, something like `dir:///foo/bar/baz`? It could be treated similarly to how the Vault datasource works, in terms of specifying sub-keys:

Let's say you had a directory like:

```
/tmp/data
         /foo.json
         /bar.yaml
```

So, you could have a template like this:
_(in.tmpl)_
```go
{{ (datasource "mydir" "foo").whatever }}
{{ (datasource "mydir" "bar").something }}
```

```console
$ gomplate -d mydir=dir:///tmp/data -i in.tmpl
```

Does that make sense?

Either way - can you work on the input/output-dir and datasource dirs as two separate PRs?

rhuss:
@hairyhenderson yeah, have seen the new options also in the README and was confused a bit that this was not yet release (but already documented). Took me a bit ;-) Having options for specfiying in and out files is a very good thing because then in an Docker environment you don't need any shell for piping things into gomplate and redirect it to files. As much as I love this as part of the Unix philosophy, this makes it impossible to use a blank base image (scratch).

If gomplate could do all the file handling, then it would be even a better fit for configuration pre processing in an orchestration environment (like in Kubernetes / OpenShift init-containers).

[...]

PR #119, opened 2017-04-20
Add --input-dir and --output-dir as options

rhuss:
- Use these options when a whole directory hierarchy needs to be processed.
- Extracted file processing logic in an extra process.go

Partial solution to #117

Bear with me, Golang is not my mother tongue (which was in fact Perl at those days, but is Java & JavaScript since some times). Looking forward to a review ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,18 @@ By default, `gomplate` will read from `Stdin` and write to `Stdout`. This behavi
 
 You can specify multiple `--file` and `--out` arguments. The same number of each much be given. This allows `gomplate` to process multiple templates _slightly_ faster than invoking `gomplate` multiple times in a row.
 
+##### `--input-dir` and `--output-dir`
+
+For processing multiple templates in a directory you can use `--input-dir` and `--output-dir` together. In this case all files in input directory will be processed as templates and the resulting files stored in `--output-dir`. The output directory will be created if it does not exist and the directory structure of the input directory will be preserved.  
+
+Example:
+
+```bash
+# Process all files in directory "templates" with the datasource given
+# and store the files with the same directory structure in "config"
+gomplate --input-dir=templates --output-dir=config --datasource config=config.yaml
+```
+
 #### `--datasource`/`-d`
 
 Add a data source in `name=URL` form. Specify multiple times to add multiple sources. The data can then be used by the [`datasource`](#datasource) function.
--- a/main.go
+++ b/main.go
@@ -2,12 +2,13 @@ package main
 
 import (
 	"io"
-	"io/ioutil"
 	"log"
 	"os"
 
 	"strings"
 	"text/template"
+
+	"errors"
 
 	"github.com/hairyhenderson/gomplate/aws"
 	"github.com/hairyhenderson/gomplate/version"
@@ -32,7 +33,6 @@ func (g *Gomplate) RunTemplate(text string, out io.Writer) {
 	if err != nil {
 		log.Fatalf("Line %q: %v\n", text, err)
 	}
-
 	if err := tmpl.Execute(out, context); err != nil {
 		panic(err)
 	}
@@ -78,43 +78,6 @@ func NewGomplate(data *Data, leftDelim, rightDelim string) *Gomplate {
 	}
 }
 
-func readInputs(input string, files []string) []string {
-	if input != "" {
-		return []string{input}
-	}
-	if len(files) == 0 {
-		files = []string{"-"}
-	}
-	ins := make([]string, len(files))
-
-	for n, filename := range files {
-		var err error
-		var inFile *os.File
-		if filename == "-" {
-			inFile = os.Stdin
-		} else {
-			inFile, err = os.Open(filename)
-			if err != nil {
-				log.Fatalf("Failed to open %s\n%v", filename, err)
-			}
-			defer inFile.Close() // nolint: errcheck
-		}
-		bytes, err := ioutil.ReadAll(inFile)
-		if err != nil {
-			log.Fatalf("Read failed for %s!\n%v\n", filename, err)
-		}
-		ins[n] = string(bytes)
-	}
-	return ins
-}
-
-func openOutFile(filename string) (out *os.File, err error) {
-	if filename == "-" {
-		return os.Stdout, nil
-	}
-	return os.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
-}
-
 func runTemplate(c *cli.Context) error {
 	defer runCleanupHooks()
 	data := NewData(c.StringSlice("datasource"), c.StringSlice("datasource-header"))
@@ -123,22 +86,51 @@ func runTemplate(c *cli.Context) error {
 
 	g := NewGomplate(data, lDelim, rDelim)
 
-	inputs := readInputs(c.String("in"), c.StringSlice("file"))
-
-	outputs := c.StringSlice("out")
-	if len(outputs) == 0 {
-		outputs = []string{"-"}
+	if err := validateInOutOptions(c); err != nil {
+		return err
 	}
 
-	for n, input := range inputs {
-		out, err := openOutFile(outputs[n])
-		if err != nil {
-			return err
+	inputDir := c.String("input-dir")
+	if inputDir != "" {
+		return processInputDir(inputDir, getOutputDir(c), g)
+	}
+
+	return processInputFiles(c.String("in"), c.StringSlice("file"), c.StringSlice("out"), g)
+}
+func getOutputDir(c *cli.Context) string {
+	out := c.String("output-dir")
+	if out != "" {
+		return out
+	}
+	return "."
+}
+
+// Called from process.go ...
+func renderTemplate(g *Gomplate, inString string, outPath string) error {
+	outFile, err := openOutFile(outPath)
+	if err != nil {
+		return err
+	}
+	// nolint: errcheck
+	defer outFile.Close()
+	g.RunTemplate(inString, outFile)
+	return nil
+}
+
+func validateInOutOptions(c *cli.Context) error {
+	if c.String("input-dir") != "" {
+		if c.String("in") != "" || len(c.StringSlice("file")) != 0 {
+			return errors.New("--input-dir can not be used together with --in or --file")
 		}
-		defer out.Close() // nolint: errcheck
-		g.RunTemplate(input, out)
 	}
-
+	if c.String("output-dir") != "" {
+		if len(c.StringSlice("out")) != 0 {
+			return errors.New("--out can not be used together with --output-dir")
+		}
+		if c.String("input-dir") == "" {
+			return errors.New("--input-dir must be set when --output-dir is set")
+		}
+	}
 	return nil
 }
 
@@ -152,15 +144,23 @@ func main() {
 	app.Flags = []cli.Flag{
 		cli.StringSliceFlag{
 			Name:  "file, f",
-			Usage: "Template file to process. Omit to use standard input (-), or use --in",
+			Usage: "Template file to process. Omit to use standard input (-), or use --in or --input-dir",
 		},
 		cli.StringFlag{
 			Name:  "in, i",
-			Usage: "Template string to process (alternative to --file)",
+			Usage: "Template string to process (alternative to --file and --input-dir)",
+		},
+		cli.StringFlag{
+			Name:  "input-dir",
+			Usage: "Directory which is examined recursively for templates (alternative to --file and --in)",
 		},
 		cli.StringSliceFlag{
 			Name:  "out, o",
 			Usage: "Output file name. Omit to use standard output (-).",
+		},
+		cli.StringFlag{
+			Name:  "output-dir",
+			Usage: "Directory to store the processed templates. Only used for --input-dir",
 		},
 		cli.StringSliceFlag{
 			Name:  "datasource, d",
--- a/process.go
+++ b/process.go
@@ -1,0 +1,123 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+// == Direct input processing ========================================
+
+func processInputFiles(stringTemplate string, input []string, output []string, g *Gomplate) error {
+	input, err := readInputs(stringTemplate, input)
+	if err != nil {
+		return err
+	}
+
+	if len(output) == 0 {
+		output = []string{"-"}
+	}
+
+	for n, input := range input {
+		if err := renderTemplate(g, input, output[n]); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// == Recursive input dir processing ======================================
+
+func processInputDir(input string, output string, g *Gomplate) error {
+	input = filepath.Clean(input)
+	output = filepath.Clean(output)
+
+	// assert tha input path exists
+	si, err := os.Stat(input)
+	if err != nil {
+		return err
+	}
+
+	// read directory
+	entries, err := ioutil.ReadDir(input)
+	if err != nil {
+		return err
+	}
+
+	// ensure output directory
+	if err = os.MkdirAll(output, si.Mode()); err != nil {
+		return err
+	}
+
+	// process or dive in again
+	for _, entry := range entries {
+		nextInPath := filepath.Join(input, entry.Name())
+		nextOutPath := filepath.Join(output, entry.Name())
+
+		if entry.IsDir() {
+			err := processInputDir(nextInPath, nextOutPath, g)
+			if err != nil {
+				return err
+			}
+		} else {
+			inString, err := readInput(nextInPath)
+			if err != nil {
+				return err
+			}
+			if err := renderTemplate(g, inString, nextOutPath); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// == File handling ================================================
+
+func readInputs(input string, files []string) ([]string, error) {
+	if input != "" {
+		return []string{input}, nil
+	}
+	if len(files) == 0 {
+		files = []string{"-"}
+	}
+	ins := make([]string, len(files))
+
+	for n, filename := range files {
+		inString, err := readInput(filename)
+		if err != nil {
+			return nil, err
+		}
+		ins[n] = inString
+	}
+	return ins, nil
+}
+
+func readInput(filename string) (string, error) {
+	var err error
+	var inFile *os.File
+	if filename == "-" {
+		inFile = os.Stdin
+	} else {
+		inFile, err = os.Open(filename)
+		if err != nil {
+			return "", fmt.Errorf("failed to open %s\n%v", filename, err)
+		}
+		// nolint: errcheck
+		defer inFile.Close()
+	}
+	bytes, err := ioutil.ReadAll(inFile)
+	if err != nil {
+		err = fmt.Errorf("read failed for %s\n%v", filename, err)
+		return "", err
+	}
+	return string(bytes), nil
+}
+
+func openOutFile(filename string) (out *os.File, err error) {
+	if filename == "-" {
+		return os.Stdout, nil
+	}
+	return os.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+}
--- a/test/files/input-dir/config.yml
+++ b/test/files/input-dir/config.yml
@@ -1,0 +1,2 @@
+one: eins
+two: zwei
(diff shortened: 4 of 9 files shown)


process.go
```diff
@@ -0,0 +1,179 @@
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// == Rendering ====================================================
+
+func renderTemplate(g *Gomplate, inString string, outPath string) error {
```
hairyhenderson: I think I'd rather this function live in `main.go`. This file seems like a pretty good place to keep all of the file I/O functions, but this one doesn't _really_ fit.

process.go
```diff
@@ -0,0 +1,179 @@
...
+		return err
+	}
+
+	outputs := c.StringSlice("out")
+	if len(outputs) == 0 {
+		outputs = []string{"-"}
+	}
+
+	for n, input := range inputs {
+		if err := renderTemplate(g, input, outputs[n]); err != nil {
```
hairyhenderson: It'd be ideal if this were done in parallel, using goroutines - I think it should be pretty straightforward...

process.go
```diff
@@ -0,0 +1,179 @@
...
+	if filename == "-" {
+		return os.Stdout, nil
+	}
+	return os.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+}
+
+func ensureDir(dir string, mode os.FileMode) error {
+	_, err := os.Stat(dir)
+	if err != nil {
+		if os.IsNotExist(err) {
```
hairyhenderson: this check is unnecessary - `os.MkdirAll` will be a no-op if the path already exists

process.go
rhuss: agreed, will move it back :)

rhuss: the only concern I have is a circular deps on the files then: main -> process -> main (or maybe I think here a bit to Javaesque). From a purity pov it should go in a dedicated render.go but thats then probably overkill.

process.go
rhuss: It probably makes sense cpu-wise, but I wonder whether io-wise it reasonable to write in parallel to disk. My gut feeling is that a serial approach might be not worse here 

But without a benchmark its hard to judge. 

Should I still change this ?  Also, this branch is only entered when using multiple --file options (not with the new --input-dir). So its not that many probably anyway.

process.go
rhuss: thanks for the heads up

process.go
rhuss: Thinking again about it, imo renderTemplate in process.go is maybe not so a bad fit as it is the final step in 'process' (so process could io + render).

test/integration/envvars.bats
```diff
@@ -11,7 +11,7 @@ load helper
 
 @test "errors with non-existant env var using .Env" {
   gomplate -i '{{.Env.FOO}}'
-  [ "$status" -eq 2 ]
+  [ "$status" -eq 1 ]
```
hairyhenderson: I'm thinking this is because `RunTemplate`'s signature has changed... I'd rather it stay the same to not break compatibility.

process.go
hairyhenderson: Well, rendering templates aren't always pure disk IO - there's often network IO that blocks for far longer. So running these in parallel could be a big win.

I didn't realize this wasn't called with `--input-dir` - I haven't had a chance to look _too_ closely yet 😉. I'd definitely rather all separate files be rendered in parallel, whether or not they come from `--file` or `--input-dir`.

process.go
hairyhenderson: > a circular deps on the files

not really an issue... to the Go compiler the files are almost irrelevant - everything in the directory is all part of the same flat namespace. Definitely a Java-esque thing - which Go isn't 😉

process.go
hairyhenderson: > imo renderTemplate in process.go is maybe not so a bad fit as it is the final step in 'process' (so process could io + render)

Figuring out which files to read and reading them is IMO a separate concern from the render step. Admittedly, `main.go` has a lot of mixed concerns, but if a new file is created I'd rather it be _fairly_ single-responsibility

main.go
```diff
@@ -26,16 +27,13 @@ type Gomplate struct {
 }
 
 // RunTemplate -
-func (g *Gomplate) RunTemplate(text string, out io.Writer) {
+func (g *Gomplate) RunTemplate(text string, out io.Writer) error {
```
hairyhenderson: This function is public API, so the signature can't change. Please remove the `error`. `panic`ing isn't the _best_ way to error out, but it's not the end of the world, and I'd _much_ rather keep API compatibility. I know of at least one user who bundles gomplate into a different tool and calls `RunTemplate` directly, so this would break them.

README.md
```diff
@@ -194,6 +194,18 @@ By default, `gomplate` will read from `Stdin` and write to `Stdout`. This behavi
 
 You can specify multiple `--file` and `--out` arguments. The same number of each much be given. This allows `gomplate` to process multiple templates _slightly_ faster than invoking `gomplate` multiple times in a row.
 
+##### `--input-dir` and `--output-dir`
+
+For processing multiple templates in a directory you can use `--input-dir` and `--output-dir` together. In this case all files in input directory will be processed as templates and the resulting files stored in `--output-dir` (which must exist). The directory structure will be preserved.
```
hairyhenderson: > `(which must exist)`

I don't think that's true, is it?

test/files/input-dir/config.yml
```diff
@@ -0,0 +1,2 @@
+one: eins
+two: zwei
```
hairyhenderson: missing EOL here, no big deal though

main.go
```diff
@@ -123,22 +86,45 @@ func runTemplate(c *cli.Context) error {
...
+	outFile, err := openOutFile(outPath)
+	if err != nil {
+		return err
 	}
+	defer checkClose(outFile, &err)
+	g.RunTemplate(inString, outFile)
+	return nil
+}
 
+func validateInOutOptions(c *cli.Context) error {
```
hairyhenderson: While this is totally necessary, I really don't like that this has to exist... I've been thinking of migrating to `github.com/jawher/mow.cli` for a while and maybe it's time to do that. It can handle option exclusion pretty well, so this validation would be unnecessary. Maybe I'll work on migrating to `mow.cli` in the next little while.

hairyhenderson:
Thanks for this, @rhuss - and sorry for the rough review so far - I want to make sure this feature is implemented well. I hope you're not taking any of my nitpicks and criticisms personally - I really appreciate the time you've put into this so far! 😄

Thinking about this further, I think I'd like `--output-dir` to be _optional_ if `--input-dir` is provided. Essentially, it would default to the current working directory (`.`) in that case.

I think I'm also going to migrate the commandline flag processing to `mow.cli` to make the option validation a little easier. It'll be a pretty disruptive PR though, so it'll require some careful merging - sorry in advance!

process.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	return nil
+}
+
+// == Recursive input dir processing ======================================
+
+func processInputDir(c *cli.Context, g *Gomplate) error {
+	inputDir := c.String("input-dir")
+	outDir := c.String("output-dir")
+	if _, err := assertDir(outDir); err != nil {
```
rhuss: @hairyhenderson here is the assertion that the `outDir` must exist. I'm not sure if should we require this, but I think its common practice to require the top-level directory to exists. Otherwise typos in the usage can easily slip through leading to unexpected results.

README.md
rhuss: See [here](https://github.com/hairyhenderson/gomplate/pull/119/files#r112611942) where it is asserted

rhuss:
> Thinking about this further, I think I'd like --output-dir to be optional if --input-dir is provided. Essentially, it would default to the current working directory (.) in that case.

ok, I can implement it that way.

rhuss:
Just added parallel file processing (both for the --file and the --input-dir case). I also did a small stupid benchmark with 1000 files a 40k each and 1000 template instructions (on an 16 GB Mac with 8 cores)

For the serial implementation I get:

```bash
time bin/gomplate --input-dir test/t/in --output-dir test/t/out --datasource config=test/t/config.yml

322.71s user 66.78s system 152% cpu 4:14.92 total
```

For the parallel implementation:

```
time bin/gomplate --input-dir test/t/in --output-dir test/t/out --datasource config=test/t/config.yml

797.23s user 151.41s system 734% cpu 2:09.22 total
```

So roughly a factor two for the total time (although more overhead for io and context switching I guess, there are 10k go routines created in this sample). Not so bad.

test/files/input-dir/config.yml
rhuss: done

main.go
rhuss: ok, let me know when and from where (branch) to rebase when you added it.

rhuss:
Addressed all review comments so far. wrt. to the parallel processing: I had to introduce some locking around the cache access when processing the Datasource. Please review this as I have quite some general experience with threading but not with shared memory when using goroutines.

rhuss:
all green ;-) like the integration test suite, need to add some for --input-dir, too ...

process.go
hairyhenderson: Hrm... 🤔 I think I'd rather create `outDir` if it doesn't exist. `gomplate` is usually used in scripts and typos would be caught pretty quickly, by virtue of the files not existing in the right location. Besides, `--out` will create the output if it doesn't exist, so it makes sense (to me anyway) that `--out-dir` would create the directory too.

Also it makes the code simpler, which is a good thing 😉

hairyhenderson:
> So roughly a factor two for the total time (although more overhead for io and context switching I guess, there are 10k go routines created in this sample). Not so bad.

Interesting! I think it would be pretty unusual to be processing directories so large, but it's 👍 anyway.

README.md
hairyhenderson: ok, commented about that in-line in `process.go`

main.go
hairyhenderson: I _think_ I might need to take this back. I started working on it in #122 but it seems to have trouble with taking `-` as an argument for `--file` or `--out`. I think it maybe makes sense to keep this code in for now and not block this PR on the other one.

data.go
```diff
@@ -174,49 +183,70 @@ func (d *Data) Datasource(alias string, args ...string) map[string]interface{} {
...
-	if d.cache == nil {
-		d.cache = make(map[string][]byte)
-	}
-	cacheKey := source.Alias
-	for _, v := range args {
-		cacheKey += v
-	}
-	cached, ok := d.cache[cacheKey]
-	if ok {
+func (d *Data) ReadSource(fs vfs.Filesystem, source *Source, args ...string) (map[string]interface{}, error) {
```
hairyhenderson: this function is public API - can you find a way to accomplish this without changing the signature?

hairyhenderson:
> wrt. to the parallel processing: I had to introduce some locking around the cache access when processing the Datasource. Please review this as I have quite some general experience with threading but not with shared memory when using goroutines.

Ok, just had a deeper look at the code, and I was a little confused about the cache lock at first, but now I understand why it's necessary.

I think I'm going to change my mind (sorry!) - let's _not_ parallelize rendering. At this point it feels like premature optimization and I'm not sure I'm comfortable with the extra complexity in the code that it introduces...

data.go
rhuss: I changed it since it was necessary to add the creation of the JSON and YAML documents in the mutex protected section when doing parallel processing. Also, I think it makes sense to include the (potentially costly) deserialized objects into the cache so that this conversion needs to be called only once (and not for every call to the datasource). I don't consider this optimisation as premature since the call to the datasource can happen quite often also when doing only a single call.

I suggest adding a second method `ReadSourceAsObjects` (could be private, too) which is used by DataSource and leave an (unoptimised) `ReadSource` with the original signature. Then we could decide whether to increase the cache to hold both kinds of objects (with the cost of increasing memory usage) or we make `ReadSource` to not use a cache.

wdyt ?

rhuss:
@hairyhenderson what about the following suggestion: We keep the current login without synchronising but add an option `--parallel` which then would switch to the threaded version. This flag could be marked as experimental.

I think it should be possible to keep the current API and clearly separate those functions without much duplication.

Please let me know whether I should give it a try. Otherwise, I revert the code again (but let me know whether you want the cache optimisation mentioned above)

process.go
rhuss: ok

hairyhenderson:
> Please let me know whether I should give it a try. Otherwise, I revert the code again (but let me know whether you want the cache optimisation mentioned above)

Let's just revert the code for now, including the cache changes. I'd rather keep this PR smaller and focused solely on `--input-dir` and `--output-dir`.

data.go
hairyhenderson: > Also, I think it makes sense to include the (potentially costly) deserialized objects into the cache so that this conversion needs to be called only once (and not for every call to the datasource). I don't consider this optimisation as premature since the call to the datasource can happen quite often also when doing only a single call.

If you're using the same datasource multiple times in a template, it makes sense to assign its output to a variable:

```go
{{ $config := datasource "config" }}
hostname: {{ $config.hostname }}
port: {{ $config.port }}
...
```

So datasources should almost never be called multiple times in a template. It's non-obvious if someone is new to gomplate and go templates in general, but maybe that could be mitigated with documentation for now... Ultimately it would be nice to have some sort of subcommand that could lint the template, but that's a long way off 😉

process.go
```diff
@@ -0,0 +1,183 @@
...
+	nr := processDir(inputDir, outDir, g, results, 0)
+	return waitAndEvaluateResults(results, nr)
+}
+
+func processDir(inPath string, outPath string, g *Gomplate, results chan *renderResult, nr int) int {
+	inPath = filepath.Clean(inPath)
+	outPath = filepath.Clean(outPath)
+
+	// assert tha input path exists
+	si, err := assertDir(inPath)
```
hairyhenderson: I think this can be avoided (and `assertDir` removed) if you just call `ioutil.ReadDir` before `os.MkdirAll`.

rhuss: From the documentation its not clear to me whether ReadDir fails properly (e.g. with a proper error message). Also the file status `si`is needed so that the output dir will have the same file mode (btw, currently the generated files all have fixed 0644, but shouldn't the have the same mode as the original template ?)

I could inline the method since it is used now only once. Will rename it to `statDir` at least.

rhuss:
Reworked again and moved the optional parallel processing to #123 which is based on this PR. Think it still makes sens, and since it was quite a bit of work and testing I didn't want to simply remove it.

rhuss:
let me know whether i should do a git squash (so weed out the threading meander on this PR)

process.go
hairyhenderson: Just a tip - the example in https://golang.org/pkg/io/ioutil/#ReadDir is editable and runnable to help verify these sorts of things.

In particular, this code:

```go
package main

import (
	"io/ioutil"
	"log"
)

func main() {
	_, err := ioutil.ReadDir("foo")
	if err != nil {
		log.Fatal(err)
	}
}
```

returns this error:
```
2009/11/10 23:00:00 open foo: No such file or directory
```

I think that's plenty clear enough 🙂

hairyhenderson: > Also the file status `si` is needed so that the output dir will have the same file mode

That makes sense. :+1:

> (btw, currently the generated files all have fixed 0644, but shouldn't the have the same mode as the original template ?)

Yup - good point. But not in this PR 😉

hairyhenderson:
> let me know whether i should do a git squash

Yes please!

main.go
```diff
@@ -123,22 +86,50 @@ func runTemplate(c *cli.Context) error {
...
-		defer out.Close() // nolint: errcheck
-		g.RunTemplate(input, out)
+	return processInputFiles(c.String("in"), c.StringSlice("file"), c.StringSlice("out"), g)
+}
+func getOutputDir(c *cli.Context) string {
+	out := c.String("output-dir")
+	if out != "" {
+		return out
 	}
+	return "."
```
hairyhenderson: This could just be accomplished by defaulting to `.` in the `StringFlag`, though that'd change the logic for `validateInOutOptions`...

process.go
```diff
@@ -0,0 +1,142 @@
...
+		return err
+	}
+
+	// ensure output directory
+	if err = os.MkdirAll(output, si.Mode()); err != nil {
+		return err
+	}
+
+	// read directory
+	entries, err := ioutil.ReadDir(input)
```
hairyhenderson: I still want this put above the `MkdirAll` call. If something goes wrong reading it, I don't want the output dir to be left behind.

process.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// == File handling ================================================
+
+func statDir(dir string) (os.FileInfo, error) {
+	si, err := os.Stat(dir)
+	if err != nil {
+		return nil, err
+	}
+	if !si.IsDir() {
```
hairyhenderson: This isn't necessary - `ioutil.ReadDir` will error if given a non-directory. If you remove this block, this function can disappear and you can just call `os.Stat` directly on ln39

process.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func openOutFile(filename string) (out *os.File, err error) {
+	if filename == "-" {
+		return os.Stdout, nil
+	}
+	return os.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+}
+
+func checkClose(c io.Closer, err *error) {
```
hairyhenderson: I'm not too clear what the point of this function is... Why not simply `defer file.Close()`? Nothing will be done with `err` so it doesn't seem too useful to assign it...

main.go
rhuss: I tried that, but this doesn't work in the validation logic, which checks for whether the outputDir is set. if it has a default value, then it is always set. You then also can't distinguish between people setting "." explicitely (and so switching on the ferature) or whether it comes from the default.

Dont think its so bad here.

process.go
rhuss: Hmm, if we want to make this really atomar, then we should also check the whole traversal to have not partially filled (and remember which directories we created and which were already there).

But I can move that up (which however makes it a bit harder to refactor of course as this method is already close to the cyclomatic complexity limit of 10)

Please be sure whether you really want this up (don't want to make many turns yet)

process.go
rhuss: ok

process.go
rhuss: It's not me, its the gometalinter which croaks if the error is not checked  ;-).

Well, if `*error` is nil then it's set with the error from the `Close()` so it makes a _little_ sense.

process.go
rhuss: moved it.

rhuss:
squashed it

process.go
hairyhenderson: Just add `// nolint: errcheck` on the line to shut `gometalinter` up.



> Well, if *error is nil then it's set with the error from the Close() so it makes a little sense.

Right, but after it's set, nothing actually _reads_ that error... 😉

main.go
hairyhenderson: > You then also can't distinguish between people setting "." explicitely (and so switching on the ferature) or whether it comes from the default.

Not _really_ sure that this matters - the feature should be dependent on `--input-dir`, no?

process.go
rhuss: actually you are right, but the calling method could initialize `err` with nil and then always return `err` (which in case of a Close() error would be filled before returning, otherwise it stays nil). That was missing.

But I agree that its an academic question, will remove that.

main.go
rhuss: How would you then error on

* when `--output-dir="."` and `--out="parsed-file.txt"` is provided
* when `--output-dir="."` is provided and no `--input-dir="."` is given

Of course I can remove these checks (or make them apply only when output-dir is not "." but that inconsistent imo).

Again, is it really that bad here ?

hairyhenderson: Hmm - good points. There would be ways to apply this, but I see now that the logic would probably be uglier - let's just leave this as-is then 🙂

process.go
hairyhenderson: thanks 🙂

hairyhenderson:
Ok! Thanks for your patience with all this - I have just one more request, then it'll be ready to go 😉 - can you write some integration tests for this? Just add some tests to `test/integration/basic.bats`, using the other tests as examples...

rhuss:
test added

hairyhenderson:
@rhuss thank you! This is ready to merge now - thanks again for your patience with this 🙂

PR #122, opened 2017-04-21
Switching commandline arg parser to jawher/mow.cli

hairyhenderson:
A few advantages to this (IMO):
- the generated usage text is a little nicer
- specifying options which are mutually exclusive is much easier


hairyhenderson:
> specifying options which are mutually exclusive is much easier

Ha! not so much. Turns out Cobra's much better. See #128

PR #123, opened 2017-04-23
Support for parallel processing with --parallel

rhuss:
This PR adds an option `--parallel` or `-p` to allow parallel template processing. This is based on PR #119 and differs only in the commit https://github.com/rhuss/gomplate/commit/964af6fd864c6ee950344b5a50eed423bc4b11ad

By default parallel processing is disabled.


hairyhenderson:
Thanks @rhuss!

I'm still not entirely sure we _need_ the extra complexity that this adds. I feel bad that I suggested you go through the work of implementing this, just to change my mind afterwards 😞.

After #119 is merged and this is re-based, I'll re-visit this.

rhuss:
Rebased

PR #125, opened 2017-04-24
Add support for --chown <uid>:<gid> 

rhuss:
The parameter can be given in the form <uid>:<gid> but also username and
group names are supported _if_ the os supports the lookup.
The safest way is to use numeric user ids.

The use case for this feature is when the processed templates are expected
to belong to a certain user/group. E.g. Wildfly Docker images, a prominent Java
EE application server write into the configuration files, too but runs with a non-root uid
and so needs the proper access. Depending on the uid under which the server
run and the files has been created, this has to be tuned.

Especially in a Docker context this might be helpful since Docker does not
support changing the ownership with ADD or COPY (its always added as root).
In order to avoid some postprocessing of the generated files (which might even
not be possible e.g. when running from a scratch image without any installed
shell), this options proves to be quite useful and add some flexibility
(on the cost of an additional feature, though).

This PR is based on #119 and add the commit https://github.com/rhuss/gomplate/commit/75a28b10476823ada27e60cb7f6b9333a88474e6


hairyhenderson:
Thanks again, @rhuss!

Wouldn't it work to just run `gomplate` as the target user?

In a non-Docker context, it should work to instead just set the uid/gid to the same as the input file (assuming the input belongs to a different user, and assuming the `gomplate` process is running as a user privileged enough to set that).

I'm thinking this option is straying a little too far outside the scope of what I want `gomplate` to be responsible for (though, admittedly, I've never actually defined that 😉)...

rhuss:
Right, it's about a Docker context, right, where you don't always have influence on the user which executes your container.

But you are probably right, it's add too much complexity for little value (especially because this feature is also OS dependent and would need to be tested cross platform).

hairyhenderson:
👍

rhuss:
@hairyhenderson btw, wonder whether original file permission shouldn't be respected or whether it always ok to have 0644 on the files created as it is done [here](https://github.com/rhuss/gomplate/blob/0d2eb6bb04761de8ebdeffb815c4e1979e2f29c4/process.go#L159-L159) I think this could be easily fixed.

hairyhenderson:
@rhuss definitely I'd like original file permissions to be respected. Originally `gomplate` didn't even deal with files (just stdin/stdout), so that's one thing that I didn't consider. I'd appreciate an issue for that, and I'd certainly accept a PR for that, if you have the time 🙂 - otherwise I'd be happy to deal with it myself.

PR #140, opened 2017-05-18
Add "replace" function and documentation

jen20:
This pull request adds a `replace` function which takes three arguments - an original string, the substring to replace, and the string with which to replace it. This is of particular use when generating node names from IP addresses where the node name may not contain "." characters.

The `strings.Replace` function in the Go standard library seems like it would be better named `strings.ReplaceN`, so we wrap it to replace all occurences of the string. If necessary, a gomplate function named `replaceN` could be introduced calling `strings.Replace` directly.

### DIFF
--- a/stringfunc.go
+++ b/stringfunc.go
@@ -1,0 +1,10 @@
+package main
+
+import "strings"
+
+// stringFunc - string manipulation function wrappers
+type stringFunc struct{}
+
+func (t stringFunc) replaceAll(old, new, s string) string {
+	return strings.Replace(s, old, new, -1)
+}
--- a/stringfunc_test.go
+++ b/stringfunc_test.go
@@ -1,0 +1,16 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestReplaceAll(t *testing.T) {
+	sf := &stringFunc{}
+
+	assert.Equal(t, "Replaced",
+		sf.replaceAll("Orig", "Replaced", "Orig"))
+	assert.Equal(t, "ReplacedReplaced",
+		sf.replaceAll("Orig", "Replaced", "OrigOrig"))
+}
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ Gomplate is an alternative that will let you process templates which also includ
 		- [`slice`](#slice)
 		- [`split`](#split)
 		- [`splitN`](#splitn)
+		- [`replaceAll`](#replaceAll)
 		- [`title`](#title)
 		- [`toLower`](#tolower)
 		- [`toUpper`](#toupper)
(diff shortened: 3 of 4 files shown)


hairyhenderson:
Hi @jen20 - thanks for the PR! This is awesome - I was thinking _just the other day_ that I needed a `replace` function 😉

Looks like the CI build is failing - should be pretty simple to fix, it seems you just need to run `go fmt` and fix up the imports - see https://circleci.com/gh/hairyhenderson/gomplate/452 for details.

stringfunc.go
```diff
@@ -0,0 +1,10 @@
+package main
+
+import "strings"
+
+// StringFunc - string manipulation function wrappers
+type StringFunc struct {}
+
+func (t StringFunc) Replace(s, old, new string) string {
```
hairyhenderson: > If necessary, a gomplate function named `replaceN` could be introduced calling `strings.Replace` directly.

To avoid the future name-shuffling (because I think exposing the full `strings.Replace` is inevitable) I think I'd prefer to name _this_ function `replaceAll`.

And as an aside, can you lower-case the `r` just so that this isn't public API right away? I feel a bit boxed in with all the exported funcs I've written to date, and I'd like to stop that and make new ones non-exported _for now_.

hairyhenderson: Can you move the `s` arg to the end of the signature? Like this:

```go
func (t StringFunc) replaceAll(old, new, s string) string {
```

Then it can be used to also support [pipelines](https://golang.org/pkg/text/template/#hdr-Pipelines):

```go
Like this:
{{ replaceAll "a" "o" "banana" }}
or this:
{{ "banana" | replaceAll "a" "o" }}
```

A number of other gomplate built-in funcs support this usage, so it'd fit in nicely.

jen20: I think all of this has been addressed with the latest push - I'll wait to watch CI complete now I'm not in a plane ;-)

stringfunc_test.go
```diff
@@ -0,0 +1,16 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestReplace(t *testing.T) {
```
hairyhenderson: Should be renamed to `TestReplaceAll`

jen20:
The last one should now be addressed. Do you have a particular schedule you are following for release builds?

hairyhenderson:
Thanks again, @jen20! I don't have a particular schedule - I should be able to get 1.7.0 out within the next day or so.

jen20:
Great - thanks for merging!

hairyhenderson:
> I should be able to get 1.7.0 out within the next day or so.

well, better late than never 😬... 1.7.0 is out now!

PR #141, opened 2017-05-18
Add `solaris-amd64` build target

jen20:
It would be great if there was an official release for `solaris/amd64` which is the platform I'm using primarily - this PR adds it as a target in `Makefile`, which as far as I can see is the only change which needs making?


hairyhenderson:
Hey, @jen20 - thanks for another great PR!

This should be the only change you need. Some platforms also get a `slim` variant, compressed with [UPX](https://upx.github.io), but as far as I can tell, UPX doesn't support Solaris, so that doesn't apply in this case.

Issue #145, opened 2017-05-22
Add `go-sockaddr` functions


jen20:
This is an issues to discuss the acceptability and design for a large feature which I think would be useful in `gomplate` - the ability to execute [go-sockaddr][gosockaddr] pipelines.

The go-sockaddr library was written primarily by @sean- for use in HashiCorp runtime tools to allow for evaluation of network interfaces against heuristics to avoid hard-coding interface names, IP addresses and so forth during configuration. Consul and Nomad (and I believe soon Vault) all support the templates in configuration fields such as `bind`.

Templates take the form of a Go function pipeline, and provide functionality such as finding an IP address that:

- is attached to a default route (`GetDefaultInterfaces()`)
- is contained within a CIDR block (`IfByNetwork()`)
- is an RFC1918 address (`IfByRFC("1918")`)
- is ordered (`OrderedIfAddrBy(args)` where args includes `AscIfType`, `AscNetworkSize`)
- excludes all IPv6 addresses (`IfByType("^(IPv4)$")`)
- is larger than a /32 (`IfByMaskSize(32)`)
- is not on a down interface (`ExcludeIfs("flags", "down")`)
- preferences an IPv6 address over an IPv4 address (`SortIfByType()` + `ReverseIfAddrs()`); and
- excludes any IP in RFC6890 address (`IfByRFC("6890")`)

I believe this would be a fairly straightforward lift-and-shift to support all of the functions supported by `go-sockaddr` in `gomplate`, and it would make rendering templates with networking aspects more flexible. Two possible approaches seem to exist:

1. Add the `go-sockaddr`-supplied functions to `gomplate` directly
1. Add a function named something like `sockaddr` which takes a template to render as a parameter.

I'm happy to take a look at doing this work, but want to confirm that the feature set is of interest prior to doing so.

[gosockaddr]: https://github.com/hashicorp/go-sockaddr

hairyhenderson:
Thanks @jen20 - on the surface, this looks like a great feature. Hadn't run into `go-sockaddr` until now, and I know I've run into situations that I'm sure would've been more elegantly solved with `go-sockaddr`. So yes, I'm interested in this.

Since this seems like it'd pull in a huge number of new functions, I want to make sure to design this in a way that'll make sense and won't paint us into a corner later.

> Two possible approaches seem to exist:
> 
> 1. Add the `go-sockaddr`-supplied functions to gomplate directly
[...]

PR #148, opened 2017-05-24
Namespacing the aws funcs

hairyhenderson:
This is an offshoot of #145 to try and figure out the best way to namespace built-in functions.

Ideally, I'd like to be able to use the existing function names, and simply move them under a namespace. So, instead of `{{ ec2region }}`, I'd like `{{ aws.ec2region }}`.

### DIFF
--- a/aws/ec2info_test.go
+++ b/aws/ec2info_test.go
@@ -8,26 +8,6 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-// test doubles
-type DummyInstanceDescriber struct {
-	tags []*ec2.Tag
-}
-
-func (d DummyInstanceDescriber) DescribeInstances(*ec2.DescribeInstancesInput) (*ec2.DescribeInstancesOutput, error) {
-	output := &ec2.DescribeInstancesOutput{
-		Reservations: []*ec2.Reservation{
-			{
-				Instances: []*ec2.Instance{
-					{
-						Tags: d.tags,
-					},
-				},
-			},
-		},
-	}
-	return output, nil
-}
-
 func TestTag_MissingKey(t *testing.T) {
 	server, ec2meta := MockServer(200, `"i-1234"`)
 	defer server.Close()
--- a/aws/testutils.go
+++ b/aws/testutils.go
@@ -5,6 +5,8 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
 )
 
 // MockServer -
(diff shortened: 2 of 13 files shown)


hairyhenderson:
The problem with this approach, as it stands right now, is that the function has to be exported to be visible, so it changes to `{{ aws.EC2Region }}`. I'm not a big fan of the restriction, though I'll live with it if I have to... ¯\\\_(ツ)_/¯

aws/aws.go
```diff
@@ -0,0 +1,21 @@
+package aws
+
+// NS -
+func NS() *awsFuncs {
+	if af == nil {
```
sean-: This conditional should be replaced with a call to [`sync.Once`](https://golang.org/pkg/sync/#Once) or initialized in an [`init()`](https://github.com/hashicorp/go-sockaddr/blob/master/template/template.go#L27-L102) function.  `sync.Once` has the advantage of you only pay for what you use, but `init()` is a bit more simple to grok and idiomatic.

aws/aws.go
```diff
@@ -0,0 +1,21 @@
...
+	return af
+}
+
+var af *awsFuncs
+var meta *Ec2Meta
+
+type awsFuncs struct{}
+
+func (a *awsFuncs) EC2Region(def ...string) string {
+	if meta == nil {
```
sean-: Ditto this conditional.

aws/aws.go
hairyhenderson: thanks, this seemed inelegant, but I didn't want to use `init()` - I'll replace with `sync.Once` 🙂

hairyhenderson:
I've been doing some digging around the upper/lower-case function name problem, and I found some discussion of how they did it in [hugo](https://gohugo.io) - https://github.com/spf13/hugo/issues/3042

It looks like basically the same approach was tried initially but in the end they seem to have gone with something more complex. I'm still wrapping my head around what exactly was done - most of the magic seems to be in https://github.com/spf13/hugo/tree/master/tpl

hairyhenderson:
Ok, seems I totally misunderstood what's going on in hugo - they _don't_ seem to support lower-cased function names in namespaces. What I was looking at was aliases, which are in the global namespace.

We're stuck with upper-cased function names. NBD. I'll move on with life now 😛

hairyhenderson:
I think this is ready to go... I've split out the docs to reflect the namespacing, and I'm not _super_ happy with the way the docs flow, but it'll have to do for now!

@sean- @jen20 - I think this approach will work for #145 - WDYT?

sean-:
@hairyhenderson Are you good with this approach now?

hairyhenderson:
@sean- yeah, it seems to be fine. There's probably further refinements that could be made to the specifics, but as a general approach I'm happy with it for now.

PR #149, opened 2017-05-24
Moving docs out of README

hairyhenderson:


hairyhenderson:
While I was at it, I figured I'd deploy the docs with Netlify at https://gomplate.hairyhenderson.ca

jen20:
Looks great! One thing I was thinking is that there should be a man page, I think there are good ways to convert Markdown into `roff` for this. Any thoughts?

hairyhenderson:
@jen20 🤔 It wouldn't be a _bad_ idea, though the markdown would probably need to be reformatted significantly. Also it'd only make sense to distribute it with the Homebrew and Alpine packages...

jen20:
Possibly (for ordering) - though I'm expecting `gomplate` to make its way into pkgsrc for SmartOS and various BSDs at some point (when I get round to submitting it), and hopefully at some point it will be packaged for Ubuntu and RPM etc.

hairyhenderson:
> I'm expecting gomplate to make its way into pkgsrc for SmartOS and various BSDs at some point (when I get round to submitting it)

😍 that'd be awesome!

Yup I'm certainly open to getting `gomplate` packaged for _all the distros_. Realistically I can't maintain all of those myself 😉.

I've logged #151 to track man page creation.

PR #175, opened 2017-07-30
Remove VFS argument from ReadSource which isn't used

stuart-c:
ReadSource is passed a VFS argument (set from source.FS) as the first argument, but this is never used. Instead source is passed so the readFile proc uses that instead (and directly accesses source.FS)


hairyhenderson:
Thanks for this! This looks good.

However, this changes public API, so I'm going to hold on to this until other major API changes need to be made (which I think will be necessary for your Vault PRs in motion right now).

PR #176, opened 2017-07-30
Switch to using offical Vault Go client

stuart-c:


hairyhenderson:
@stuart-c thanks for this - looks like #177 duplicates this?

Is there a particular reason for this PR? The reason I don't use the official client is so as not to pull in all the extra unnecessary dependencies.

This PR causes the binary size to increase by almost 50% (~7.5MB).

stuart-c:
Replaced by #177 which has a smaller size impact

Issue #173, opened 2017-07-27
Consul support


stuart-c:
Would support for substitutions based on values stored in the key/value store within Consul be considered (optionally with authentication via Vault)?

hairyhenderson:
@stuart-c Thanks for the issue! Yes, absolutely. I'd love to have Consul and other k/v stores as datasources, and have started writing this a few times over the last months 😉. A while back I had a look at https://github.com/docker/libkv as a potential way to make this simple, but it may be overkill.

As for authentication via Vault (I'm assuming you're talking about https://www.vaultproject.io/docs/secrets/consul/?) - it could be useful but I think I'd like to get unauthenticated Consul support in first.

hairyhenderson:
Implemented by #178 🎉

PR #177, opened 2017-07-31
Switch to using official Go Vault client

stuart-c:


hairyhenderson:
@stuart-c thanks for this. Is there a particular reason for this PR?

The reason I don't use the official client is so as not to pull in all the extra unnecessary dependencies. This particular PR is leaner than #176 - it only increases the binary by 3%. But, it's still an increase and I'd rather have a good reason to make the change.

Thanks!

stuart-c:
As you saw this is a duplicate of #176 but with a smaller overhead.

I'm currently working on trying to add a few bits which this fits into:

* #173 adding libkv - some of the extra dependencies will be shared, so the overhead is actually a bit less
* Fixing a few bugs in the current implementation (mostly the _FILE option not working)
* Adding write support for Vault to allow dynamic secrets (easier with the client instead of having to re-implement)
* Full Vault option support (e.g. SSL options, etc.) which work out of the box with the client
* Use of Vault for libkv authentication

hairyhenderson:
Ok, so I guess this is more a pre-requisite to things you have planned further down the line. That sounds OK to me. Can you close #176 if this supersedes it?

I'll review this in closer detail soon. Again, thanks for the PR!

vault.go
```diff
@@ -0,0 +1,234 @@
+package main
```
hairyhenderson: Can you move this into the `vault` subdirectory/package?

vault.go
```diff
@@ -0,0 +1,234 @@
...
+func (v *Vault) Logout() {
+}
+
+func (v *Vault) Read(path string) ([]byte, error) {
+	secret, err := v.client.Logical().Read(path)
+	if err != nil {
+		return nil, err
+	}
+
+	return jsonutil.EncodeJSON(secret.Data)
```
hairyhenderson: I'm not sure about the extra expense of `jsonutil` being pulled in - please replace with just `json.NewEncoder(...).Encode(secret.Data)`

env.go
```diff
@@ -17,3 +22,26 @@ func (e *Env) Getenv(key string, def ...string) string {
 
 	return val
 }
+
+// GetenvFile - Also checks file
+func (e *Env) GetenvFile(fs vfs.Filesystem, key, def string) string {
```
hairyhenderson: This isn't really usable as a template function as-is, which I think would be quite useful. For it to be usable as a template function, it'd need to have a signature like:

```go
func (e *Env) GetenvFile(key string, def ...string) string {
```

Since we can't pass `vfs.Filesystem`s in through templates, and `def` needs to be optional.

What could be useful would be to have a "template" version of this with the above signature, and an "internal" version with this syntax.

All that being said, I think an addition like this may benefit from being in a separate PR. It may also be useful to have this moved to a new `env` package so it can be used by other packages (to avoid circular references).

hairyhenderson:
@stuart-c I would also like to see some unit tests for the remaining code in `vault.go`

env.go
hairyhenderson: > All that being said, I think an addition like this may benefit from being in a separate PR. It may also be useful to have this moved to a new `env` package so it can be used by other packages (to avoid circular references).

On second thought, I'll take care of this piece myself - don't worry about a separate PR. I'll get you to rebase after I've moved to a new `env` package.

vault.go
stuart-c: I did actually do something similar previously but jsonutil is pulled in anyway. I can do it either way...

hairyhenderson: I'd rather use the standard library method. I'll be migrating from using glide to dep at some point soon, and with dep I'd like to prune before building, so reducing dependencies is helpful.

env.go
stuart-c: Are you going to add the GetenvFile func too?

hairyhenderson: @stuart-c yep - working on it right now actually

hairyhenderson: Ok, it's in https://github.com/hairyhenderson/gomplate/pull/181 - I made a change to the approach - instead of having a separate function, the functionality is built-in to the existing `Getenv` function. This makes it easier to use in a template context, since the template author doesn't need to know how the user is going to provide the data.

vault.go
stuart-c: Could I have a hand here as I'm struggling to get the import right...

hairyhenderson: Still struggling? Looks like it's compiling now...

stuart-c: I reverted the package change

hairyhenderson: Oh, ok. You should just be able to move this into `vault` subdirectory and change this line to `package vault`. Let me know if something isn't working right beyond that...

stuart-c: It's more the import I'm unsure of

hairyhenderson: Can you explain? Should just be `"github.com/hairyhenderson/gomplate/vault"`

stuart-c: Yep, but that vault directory is the current package rather than the new one?

hairyhenderson: I'm not sure I follow - will pull your fork and see what I can see 🙂

hairyhenderson: Ok, this worked for me.

Diff for `data.go`:
```diff
diff --git a/data.go b/data.go
index 0776b65..32486fd 100644
--- a/data.go
+++ b/data.go
@@ -14,6 +14,7 @@ import (
        "time"
 
        "github.com/blang/vfs"
+       "github.com/hairyhenderson/gomplate/vault"
 )
 
 // logFatal is defined so log.Fatal calls can be overridden for testing
@@ -83,7 +84,7 @@ type Source struct {
        Params map[string]string
        FS     vfs.Filesystem // used for file: URLs, nil otherwise
        HC     *http.Client   // used for http[s]: URLs, nil otherwise
-       VC     *Vault         // used for vault: URLs, nil otherwise
+       VC     *vault.Vault   // used for vault: URLs, nil otherwise
        Header http.Header    // used for http[s]: URLs, nil otherwise
 }
 
@@ -303,7 +304,7 @@ func readHTTP(source *Source, args ...string) ([]byte, error) {
 
 func readVault(source *Source, args ...string) ([]byte, error) {
        if source.VC == nil {
-               source.VC = NewVault()
+               source.VC = vault.NewVault()
                source.VC.Login()
                addCleanupHook(source.VC.Logout)
        }
```

Diff for `vault.go`:
```diff
diff vault.go vault/vault.go 
1c1
< package main
---
> package vault
```

stuart-c: Thanks. Managed to fix it - was some local GOPATH issue too.

stuart-c:
What is still needed for this PR?

hairyhenderson:
@stuart-c I think it's pretty much ready to go - I do want some unit tests, but I can handle that later as I would like to get this merged.

PR #190, opened 2017-08-04
Vault AWS EC2 auth

stuart-c:
This requires #177 to be merged first.

Add AWS ec2 auth support for Vault.

### DIFF
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -5,9 +5,13 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"strings"
+	"time"
 
 	"github.com/blang/vfs"
+	"github.com/hairyhenderson/gomplate/aws"
 	"github.com/hairyhenderson/gomplate/env"
+	"github.com/hairyhenderson/gomplate/typeconv"
 )
 
 // GetToken -
@@ -25,6 +29,9 @@ func (v *Vault) GetToken() string {
 		return token
 	}
 	if token := v.TokenLogin(); token != "" {
+		return token
+	}
+	if token := v.EC2Login(); token != "" {
 		return token
 	}
 	logFatal("All vault auth failed")
@@ -143,6 +150,41 @@ func (v *Vault) UserPassLogin() string {
 	}
 	if secret == nil {
 		logFatal("Empty response from UserPass logon")
+	}
+
+	return secret.Auth.ClientToken
+}
+
+// EC2Login - AWS EC2 auth backend
+func (v *Vault) EC2Login() string {
+	role := env.Getenv("VAULT_AUTH_AWS_ROLE")
+	mount := env.Getenv("VAULT_AUTH_AWS_MOUNT", "aws")
+
+	vars := map[string]interface{}{}
+
+	if role != "" {
+		vars["role"] = role
+	}
+
+	opts := aws.ClientOptions{
+		Timeout: time.Duration(typeconv.MustAtoi(os.Getenv("AWS_TIMEOUT"))) * time.Millisecond,
+	}
+
+	meta := aws.NewEc2Meta(opts)
+
+	vars["pkcs7"] = strings.Replace(strings.TrimSpace(meta.Dynamic("instance-identity/pkcs7")), "\n", "", -1)
+
+	if vars["pkcs7"] == "" {
+		return ""
+	}
+
+	path := fmt.Sprintf("auth/%s/login", mount)
+	secret, err := v.client.Logical().Write(path, vars)
+	if err != nil {
+		logFatal("AWS EC2 logon failed", err)
+	}
+	if secret == nil {
+		logFatal("Empty response from AWS EC2 logon")
 	}
 
 	return secret.Auth.ClientToken
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -172,10 +172,6 @@ func (v *Vault) EC2Login() string {
 
 	meta := aws.NewEc2Meta(opts)
 
-	if endpoint := env.Getenv("AWS_META_ENDPOINT"); endpoint != "" {
-		meta.Endpoint = endpoint
-	}
-
 	vars["pkcs7"] = strings.Replace(strings.TrimSpace(meta.Dynamic("instance-identity/pkcs7")), "\n", "", -1)
 
 	if vars["pkcs7"] == "" {
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -172,6 +172,10 @@ func (v *Vault) EC2Login() string {
 
 	meta := aws.NewEc2Meta(opts)
 
+	if endpoint := env.Getenv("AWS_META_ENDPOINT"); endpoint != "" {
+		meta.Endpoint = endpoint
+	}
+
 	vars["pkcs7"] = strings.Replace(strings.TrimSpace(meta.Dynamic("instance-identity/pkcs7")), "\n", "", -1)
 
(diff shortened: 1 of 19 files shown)


hairyhenderson:
Awesome! This'll need to be rebased now that #177 is merged.

vault/auth.go
```diff
@@ -148,6 +155,63 @@ func (v *Vault) UserPassLogin() string {
 	return secret.Auth.ClientToken
 }
 
+// EC2Login - AWS EC2 auth backend
+func (v *Vault) EC2Login() string {
+	if env.Getenv("VAULT_AUTH_AWS_METHOD", "ec2") != "ec2" {
+		return ""
+	}
+
+	if skip := env.Getenv("VAULT_AUTH_AWS_EC2_SKIP", ""); skip != "" {
```
hairyhenderson: I think if `VAULT_AUTH_AWS_METHOD` is required (i.e. _must_ be set to `ec2`, and doesn't default), then this environment variable can be dropped.

stuart-c: I was thinking of moving it last in the order so both environment variables can be removed

hairyhenderson: Isn't `VAULT_AUTH_AWS_METHOD` required to differentiate between `ec2` and `iam`? Or do you mean you'd default to `ec2`?

hairyhenderson: As far as moving it last in order, that's _probably_ fine. There are cases where I use gomplate in CI environments and pass it a `VAULT_TOKEN` var so it can auth correctly. Those CI environments are generally in EC2, so it'd be good to not get tripped up there.

vault/auth.go
```diff
@@ -148,6 +155,63 @@ func (v *Vault) UserPassLogin() string {
...
+	opts := aws.ClientOptions{}
+
+	timeout := os.Getenv("AWS_TIMEOUT")
+	if timeout != "" {
+		t, err := strconv.Atoi(timeout)
+		if err != nil {
+			logFatal("Invalid AWS_TIMEOUT value '%s' - must be an integer\n", timeout)
+		}
+
+		opts.Timeout = time.Duration(t) * time.Millisecond
```
hairyhenderson: I just exposed a new `MustAtoi` in `github.com/hairyhenderson/gomplate/typeconv`, so you can do this instead:

```go
opts.Timeout = time.Duration(typeconv.MustAtoi(os.Getenv("AWS_TIMEOUT"))) * time.Millisecond
```

stuart-c:
I'm just trying to add some integration tests.

vault/auth.go
```diff
@@ -148,6 +155,44 @@ func (v *Vault) UserPassLogin() string {
...
+	vars := map[string]interface{}{}
+
+	if role != "" {
+		vars["role"] = role
+	}
+
+	opts := aws.ClientOptions{}
+
+	timeout := os.Getenv("AWS_TIMEOUT")
+	if timeout != "" {
```
hairyhenderson: There's actually no need to have this check, because `opts.Timeout`'s default is `0` anyways. There's no harm to simply setting it to `0` explicitly.

You could make things even simpler by setting `Timeout` while creating the struct (up on ln169):

```go
opts := aws.ClientOptions{
  Timeout: time.Duration(typeconv.MustAtoi(os.Getenv("AWS_TIMEOUT"))) * time.Millisecond,
}
```

stuart-c:
@hairyhenderson could you take a look at this now?

hairyhenderson:
@stuart-c the integration test is failing:

```
not ok 33 Testing ec2 vault auth
# (in test file test/integration/datasources_vault.bats, line 134)
#   `vault auth-enable aws' failed with status 2
# Success! Data written to: secret/foo
# Error: Error making API request.
# 
# URL: POST http://127.0.0.1:8200/v1/sys/auth/aws
# Code: 400. Errors:
# 
# * unknown backend type: aws
# Success! Deleted 'secret/foo' if it existed.
# Disabled auth provider at path 'userpass' if it was enabled
# Disabled auth provider at path 'userpass2' if it was enabled
# Disabled auth provider at path 'approle' if it was enabled
# Disabled auth provider at path 'approle2' if it was enabled
# Disabled auth provider at path 'app-id' if it was enabled
# Disabled auth provider at path 'app-id2' if it was enabled
# Disabled auth provider at path 'aws' if it was enabled
# Policy 'writepol' deleted.
# Policy 'readpol' deleted.
# Successfully unmounted 'ssh' if it was mounted
```

vault/auth.go
```diff
@@ -148,6 +155,45 @@ func (v *Vault) UserPassLogin() string {
...
+	}
+
+	opts := aws.ClientOptions{
+		Timeout: time.Duration(typeconv.MustAtoi(os.Getenv("AWS_TIMEOUT"))) * time.Millisecond,
+	}
+
+	meta := aws.NewEc2Meta(opts)
+
+	if endpoint := env.Getenv("AWS_META_ENDPOINT"); endpoint != "" {
+		meta.Endpoint = endpoint
```
hairyhenderson: I think this block belongs here instead: https://github.com/hairyhenderson/gomplate/blob/master/aws/ec2meta.go#L89

vault/auth.go
```diff
@@ -148,6 +155,45 @@ func (v *Vault) UserPassLogin() string {
...
+	}
+
+	vars["pkcs7"] = strings.Replace(strings.TrimSpace(meta.Dynamic("instance-identity/pkcs7")), "\n", "", -1)
```
hairyhenderson: Why the extra `strings.Replace`? `strings.TrimSpace` will trim off `\n` characters... Or is this intended to remove all newlines in the middle of the string?

hairyhenderson:
> the integration test is failing

I wonder if the dev-mode vault doesn't support the `aws` secret backend 🤔

hairyhenderson:
> the integration test is failing

Ah - it's because the version of Vault in `hairyhenderson/gomplate-ci-build` is too old - just pushed an update in hairyhenderson/dockerfiles@7af0f95dc27048dc937505ddb07f1dcf48e98f08

hairyhenderson:
@stuart-c can you bump the Vault version in https://github.com/hairyhenderson/gomplate/blob/master/test/integration/Dockerfile#L3 so that `make test-integration-docker` succeeds?

vault/auth.go
stuart-c: Yep all embedded newlines need to be removed.

vault/auth.go
stuart-c: You mean you'd prefer the AWS_META_ENDPOINT to be set in the ec2meta file?

hairyhenderson: I'd prefer it to be read there - so that the metadata endpoint can be overridden in general, not just when using Vault EC2 auth

stuart-c:
That is now fixed, but the tests are hanging a bit further down. Not sure what's going on. Is it working for you locally?

hairyhenderson:
Looks like you accidentally committed `test/integration/aws` and ` test/integration/meta` - maybe should add those to a `.gitignore`

hairyhenderson:
> the tests are hanging a bit further down. Not sure what's going on

Maybe related to a CircleCI issue: https://status.circleci.com/incidents/1hf3n7yf8zrn?

stuart-c:
Fixed the .gitignore

stuart-c:
Looks to be a bats bug which I think I've now worked around...

So I think everything should now be good @hairyhenderson?

PR #191, opened 2017-08-04
[WIP] Vault AWS IAM auth

stuart-c:
Adds IAM auth for Vault.

Not sure this is the best way - have added a chunk of code from cli.go in vault (builtin/credential/aws) as otherwise there are far too many extra dependencies.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -99,7 +99,7 @@
 
 [[projects]]
   name = "github.com/hashicorp/vault"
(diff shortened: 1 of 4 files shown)


hairyhenderson:
@stuart-c is this still in-progress? If so it needs to be rebased and merge conflicts resolved...

hairyhenderson:
This is pretty stale at this point...

PR #202, opened 2017-09-07
Add nonce support

stuart-c:
My previous AWS auth PR unfortunately had an omission - it didn't handle nonces. As a result it isn't possible to run gomplate a second time.

This PR fixes this oversight.


vault/auth.go
```diff
@@ -159,13 +160,25 @@ func (v *Vault) UserPassLogin() string {
 func (v *Vault) EC2Login() string {
 	role := env.Getenv("VAULT_AUTH_AWS_ROLE")
 	mount := env.Getenv("VAULT_AUTH_AWS_MOUNT", "aws")
+	nonce := env.Getenv("VAULT_AUTH_AWS_NONCE")
+	output := env.Getenv("VAULT_AUTH_AWS_NONCE_OUTPUT")
+	perms := env.Getenv("VAULT_AUTH_AWS_NONCE_PERMISSION", "0666")
```
hairyhenderson: I'm not sure I'm a fan of `0666` as a default permission since it'll be group/other read/writable. IMO `0600` is a safer default.

Also is there even a need for this to be configurable? I'm not sure I really see a use-case where someone would need to save a nonce that's world-readable, and can't simply `chmod` it after `gomplate` runs...

vault/auth.go
```diff
@@ -159,13 +160,25 @@ func (v *Vault) UserPassLogin() string {
 func (v *Vault) EC2Login() string {
 	role := env.Getenv("VAULT_AUTH_AWS_ROLE")
 	mount := env.Getenv("VAULT_AUTH_AWS_MOUNT", "aws")
+	nonce := env.Getenv("VAULT_AUTH_AWS_NONCE")
+	output := env.Getenv("VAULT_AUTH_AWS_NONCE_OUTPUT")
```
hairyhenderson: I wonder if maybe naming this `VAULT_AUTH_AWS_NONCE_FILE` would be better, and making it possible to provide a nonce via a file (the case where the file exists already), or save it to that path if it's nonexistant.

hairyhenderson:
@stuart-c looks like the integration tests are failing in CI:

```
not ok 35 Testing ec2 vault auth
# (in test file test/integration/datasources_vault.bats, line 143)
#   `curl -o $tmpdir/certificate -s -f http://127.0.0.1:8081/certificate' failed with status 7
```

vault/auth.go
stuart-c: I can change the default or make it non-configurable.

vault/auth.go
stuart-c: That should already work as it should be using the env lookup with _FILE support

vault/auth.go
hairyhenderson: ok, I'd rather make it explicitly `0600` lower down, and also remove the option to configure

vault/auth.go
stuart-c: Changed

stuart-c:
Not sure why the integration test has started failing :(

vault/auth.go
stuart-c: Done

hairyhenderson:
> Not sure why the integration test has started failing

Me neither... I'll chalk it up to something timing-related, as I see it's passing now...

Thanks for this PR!

stuart-c:
Great. Any chance of a new release?

Thanks

hairyhenderson:
@stuart-c yup - I'll work on one soon!

PR #207, opened 2017-09-14
Remove trailing spaces

### LINKED ISSUE #205: 'client nonce mismatch' when using AWS auth with nonce
When using VAULT_AUTH_AWS_NONCE_OUTPUT in combination with VAULT_AUTH_AWS_NONCE_FILE, the nonce value is not used correctly, making gomplate unable to authenticate.

When no value exists, gomplate authenticates and writes out the nonce value:
```
[centos@ip-10-101-207-71 ~]$ touch nonce
[centos@ip-10-101-207-71 ~]$ export VAULT_AUTH_AWS_NONCE_FILE="$(pwd)/nonce"
[centos@ip-10-101-207-71 ~]$ export VAULT_AUTH_AWS_NONCE_OUTPUT="$(pwd)/nonce"
[centos@ip-10-101-207-71 ~]$ echo '{{(datasource "vault" "secret/test").value}}' | gomplate --datasource vault=vault://
foo
[centos@ip-10-101-207-71 ~]$ cat nonce
6890ac46-c75c-0c28-2570-f44e11d35860
```

But running the same command a second time throws an error:
```
[centos@ip-10-101-207-71 ~]$ echo '{{(datasource "vault" "secret/test").value}}' | gomplate --datasource vault=vault://
2017/09/13 13:02:08 AWS EC2 logon failedError making API request.

URL: PUT https://vault.cloud.local:443/v1/auth/aws/login
Code: 400. Errors:

* client nonce mismatch
```

It appears that this is because of the trailing newline in the file, as it works if you remove it:
```
[centos@ip-10-101-207-71 ~]$ echo -n "6890ac46-c75c-0c28-2570-f44e11d35860" > nonce
[centos@ip-10-101-207-71 ~]$ echo '{{(datasource "vault" "secret/test").value}}' | gomplate --datasource vault=vault://
foo
```

Presumably gomplate should either not include the trailing newline, or should be able to strip it when it reads the file?

stuart-c:
To fix #205

### DIFF
--- a/env/env.go
+++ b/env/env.go
@@ -3,6 +3,7 @@ package env
 import (
 	"io/ioutil"
 	"os"
+	"strings"
 
 	"github.com/blang/vfs"
 )
@@ -37,7 +38,7 @@ func getenvFile(fs vfs.Filesystem, key string) string {
 		if err != nil {
 			return ""
 		}
-		return val
+		return strings.TrimSpace(val)
 	}
 
 	return ""


hairyhenderson:
@stuart-c this is OK in theory, but is it necessary to address #205? Can't you simply remove the `\n` like mentioned in https://github.com/hairyhenderson/gomplate/issues/205#issuecomment-329346856 ?

stuart-c:
The reason I went with this approach is:

- Newlines aren't valid at the end of most if not all the various env vars we are using (e.g. CONSUL_VAULT_MOUNT)
- Many tools will add newlines to the end of files (e.g. echo '' > file will unless you remember to add -n)
- Having a newline in the produced nonce file is nice (e.g. cat works without having to remember to add the newline on the command line)

So this change would fix the nonce problem, but also fix possible future issues where people are using say VAULT_SECRET_ID_FILE with a file with trailing newline.

hairyhenderson:
Thanks for the explanation, @stuart-c. That makes sense to me!

PR #208, opened 2017-09-16
Fixes #192 - fail with unknown args

### LINKED ISSUE #192: "gomplate -in flibbit" should produce an error
I would expect that non-flag arguments would raise an error in `gomplate`.

Expected:

``` sh
$ gomplate -in flibbit
I don't know what to do with 'flibbit'
# non-zero exit code
$
```

Actual:

``` sh
$ gomplate -in flibbit
n<no newline> 
# zero exit code
$
```

I got bit by this because I typed `-in` instead of `--in` and spent a while trying to figure out what I did wrong.

FYI: `-in` can be rewritten as `--in 'n'`

drmdrew:
Fixes #192 

Test case:
```
gomplate -in flibbit
```

This PR causes gomplate to exit with a non-zero exit status and display an error message to the user:
```
Error: unknown argument 'flibbit'
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .git
 bin
 report.xml
+gomplate
--- a/main.go
+++ b/main.go
@@ -68,12 +68,7 @@ func newGomplateCmd() *cobra.Command {
 			}
 			return runTemplate(&opts)
 		},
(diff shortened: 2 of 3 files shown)


hairyhenderson:
Thanks for the bug fix @drmdrew!

Issue #218, opened 2017-10-26
Exclude option support


zealic:
Need "--exclude" option to ignore some files:

Example:
```
gomplate --input-dir=./data \
  --exclude=".*"     \ # Exclude file or directory with dot prefix
  --exclude="*.xml"  \ # Recursive exclude xml file
  --exclude="logs"   \ # Exclude logs file or directory
  --output-dir=./output
```

Refernece:
* https://golang.org/pkg/path/filepath/#Glob

hairyhenderson:
@zealic thanks for logging this! Sounds like a reasonable feature when processing a directory with many files.

zealic:
I will implement this feature.
Maybe '.gomplateignore' is better idea.

hairyhenderson:
@zealic yeah - `.gomplateignore` would be great - I'm assuming you're meaning something similar to the `.gitignore` file?

zealic:
@hairyhenderson Yes, it is.

Default use '--input-dir' root .gomplateignore file, sub folder can be inherit.
Finally, '--exclude' option add additional ignore files.

hairyhenderson:
@zealic That sounds good. I'd like the `.gomplateignore` and `--exclude` features to be submitted as separate PRs, for clarity

keydrevolta:
@zealic this may help you https://github.com/tsuru/tsuru-client/pull/25

Gman98ish:
Hi, I've not done the whole thing, just found a potential solution for the command line flag, potentially provides a starting point for the `.gomplateignore`

hairyhenderson:
`--exclude` is now available, so I'm going to close this issue. Feel free to enhance with further PRs though! 🙂

PR #225, opened 2017-11-02
Add .gomplateignore support

zealic:
For #218


hairyhenderson:
Thanks for this, @zealic! When I have a little more time I'll review this more fully, but in the meantime can you add an integration test? Have a look at the existing `.bats` files in `test/integration/` for inspiration. You can run the tests with `test-integration-docker`. I'm mostly looking for automated confirmation that the compiled binary processes `.gomplateignore` files as expected.

zealic:
OK, I will try add integration tests. @hairyhenderson

zealic:
Integration tests added, please review and merge.

Gopkg.toml
```diff
@@ -14,3 +14,7 @@
 [[constraint]]
   branch = "v2"
   name = "gopkg.in/hairyhenderson/yaml.v2"
+
+[[constraint]]
+  name = "github.com/zealic/xignore"
+  version = "0.3.2"
```
hairyhenderson: is there a reason you're pinning to this release? this shouldn't be necessary in this case - there's a tag and `dep` will understand it just fine...

docs/content/usage.md
```diff
@@ -35,6 +35,8 @@ You can specify multiple `--file` and `--out` arguments. The same number of each
 
 For processing multiple templates in a directory you can use `--input-dir` and `--output-dir` together. In this case all files in input directory will be processed as templates and the resulting files stored in `--output-dir`. The output directory will be created if it does not exist and the directory structure of the input directory will be preserved.  
 
+You can use `.gomplateignore` ignore some files, have the similar behavior to the [.gitignore](https://git-scm.com/docs/gitignore) file.
```
hairyhenderson:
```suggestion
You can use `.gomplateignore` to ignore some files in the input directory, with similar syntax and behaviour to [.gitignore](https://git-scm.com/docs/gitignore) files.
```

hairyhenderson:
Thanks for updating this, @zealic! I'll be able to take a deeper look in a few hours, hopefully.

Gopkg.toml
zealic: I think minor range version `~` is better than fixed version.
I will maintain upstream repo to keep stability

I will maintain upstream repo to keep stablelity.

## See also:
* https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#version
* https://semver.org

template.go
```diff
@@ -14,8 +14,12 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/spf13/afero"
+	"github.com/zealic/xignore"
 )
 
+// Gomplateignore ignore file name, like .gitignore
+const Gomplateignore = ".gomplateignore"
```
hairyhenderson: Let's leave this unexported for now - I don't see a reason to export this yet:

```suggestion
const gomplateignore = ".gomplateignore"
```

(obviously references below will need to change)

hairyhenderson:
Thanks @zealic! And, thanks especially for your persistence in getting this in - I know it took a very long time! 😉

Issue #116, opened 2017-04-17
Support Kubernetes ConfigMaps as datasources


rhuss:
It would be awesome if Kubernetes [ConfigMaps](https://kubernetes.io/docs/tasks/configure-pod-container/configmap/) could be used as datasources. This would be a perfect fit when gomplate is used as an [init-container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) which can be used to prepare configuration files for an application (like a `standalone.xml` for Wildfly app server).

Currently this can be done via ConfigMap backed volumes but a direct K8s support would be nicer, of course as it reduces the configuration overhead within Kubernetes.

rhuss:
The would make sense for [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) probably, too.

hairyhenderson:
Thanks @rhuss - I'm not very familiar with Kubernetes, but looking at the API, this could be fairly easy to implement using an HTTP(S) `datasource` (see https://kubernetes.io/docs/api-reference/v1.5/#read-199 for example).

One thing I'm not sure about is authentication, though #115 went in recently which allows arbitrary headers to be sent (mostly for auth purposes).

I think a good start would be to try to see if something like this works:

_(in.tmpl)_
```go
{{ (datasource "myconfigmap").data | toJSON }}
```

```console
$ gomplate --datasource myconfigmap=http://my.kube.api.address/api/v1/namespaces/{namespace}/configmaps/myconfigmap -f in.tmpl
```

It'd probably be nicer to have simplified support, but hopefully this is a start! Can you try this out and let me know if it works? Or let me know if I'm way off and this makes no sense 😉

rhuss:
Kubernetes comes with a full client API which you can use directly without going over HTTP. As this datasource is probably supposed to be used within a Kubernetes cluster, the API server can be easily used with a so called `ServiceAccount`, so there is not to worry much about authentication. 

The use case I have in mind is agnostic of the API server's address as it is running from within the cluster (not from outside where it doesn't make much sense imo). 

As seen also in the [SO answer](http://stackoverflow.com/a/33109739/207604) all connection parameters can be directly fetched from files which are installed by Kubernetes at fixed file locations within the container (so gomplate could simply pick them up there). The namespace can also be picked up directly (would have to check how, though). The only parameters left is then configmap name.

It''s not urgent, just wanted to open the issue more as reminder for a cool, unique feature  ;-)

[...]

Issue #213, opened 2017-10-10
Default values for variables


vascop:
Is there a way to use a default value when a variable isn't defined?

Something like:
`{{(ds "config").notes | default "this a default note")}}`

hairyhenderson:
@vascop Not in a general sense right now. For your specific example, that might result in an error if the `config` datasource doesn't have a `notes` key. I'd have to dig a little further into it to make sure.

Either way - this is a good idea! Thanks for logging this issue 🙂

zealic:
I think we need a new func: `hasKey`

like hugo's [isset](https://gohugo.io/functions/isset/)

```
{{ $config := (ds "config") -}}
{{ $notes := "this a default note" -}}
{{ if hasKey $confg "notes" -}}
{{ $notes := $config.notes -}}
{{ end -}}
```

hairyhenderson:
@zealic there is `has`/`conv.Has` already: https://gomplate.hairyhenderson.ca/functions/conv/#conv-has

hairyhenderson:
And actually, that's an excellent point... @vascop will `has` fix your use case? What @zealic suggested should work (except use `has` instead of `hasKey`).

vascop:
I think it'd probably work, though much more verbose since we'd have to do the whole "check if it has key and overwrite" dance instead of one-lining the default values.

hairyhenderson:
@vascop understood. It's not ideal. But it might be the only option...

The challenge is that referencing an undefined key will cause Go's `text/template` to blow up:

```
$gomplate -i '{{ (json `{"foo":"bar"}`).baz }}'
panic: template: template:1:9: executing "template" at <`{"foo":"bar"}`>: map has no entry for key "baz"

goroutine 1 [running]:
main.(*Gomplate).RunTemplate(0xc42030e030, 0x7fff5fbff54c, 0x20, 0x1e95de0, 0xc42000c018)
	/go/src/github.com/hairyhenderson/gomplate/gomplate.go:31 +0x42c
```

My impression is that we won't be able to get that _specific_ syntax to work without re-implementing Go's templating system.

What we _can_ do is have a `default` function that will provide a value if the given input is `nil` or an empty string. That's not too difficult, but it wouldn't address your specific use-case.

estahn:
👍 for `default`, e.g. `{{ .foo | default "bar" }}`.

hairyhenderson:
[...]

PR #235, opened 2017-12-02
add k8s configmap as source

josledp:
Solves #116. 
Currently it lacks integration tests. I would like to know which kind of integration test you may prefer:
1. Using minikube (https://github.com/kubernetes/minikube). Basically it will spin up a virtualmaching with an all-in-one node Kubernetes cluster (this is how I have developed this new data source)
2. Graping a valid Kubernetes API answer to emulate it (as with awssvc)

The good point about 1 is that it will also help with any other k8s datasource (Secrets for example), but it is more complex and heavy.


hairyhenderson:
@josledp Thanks! I'll look at this soon. Can you split the vendoring changes into a separate commit please?

As for integration tests, they'll need to be able to run in the CI system, within a Docker container (in CircleCI). I'm not sure what the best way to accomplish this is, though - I'm not very well-versed on k8s 🙂

josledp:
Ok, I've splitted the commit. About integration test, I will then emulate the k8s API, because circle ci will not be able to spin up a minikube.

josledp:
I've created an integration tests for k8s, but just realized that the docker container used by CircleCi does not test aws/meta/... (I've updated the Dockerfile for the "local" integration test). So now I've broken the CircleCi integration tests :)
Could you take a look and tellme how to properly configure the integration tests? (do I remove the datasource_k8s.bats? or should I update de CircleCI dockerfile?). If I understood correctly how the integration tests are done, CircleCI is not testing aws,meta and mirror datasources right?

hairyhenderson:
@josledp Thanks - I've only been able to have a brief look at this just now (sorry, busy weekend for me).

This PR adds a new function for accessing k8s, which isn't quite what I was expecting. This needs to use the existing datasource functionality, i.e: the `--datasource` commandline option, and the `datasource` function.

Maybe using the Consul datasource as a starting point will help. Have a look at: https://github.com/hairyhenderson/gomplate/blob/master/data/datasource.go#L407..L427 (it's referenced here: https://github.com/hairyhenderson/gomplate/blob/master/data/datasource.go#L52).

I also wonder if _quite so many_ dependencies are necessary... I've heard the Kubernetes API is enormous, though, so maybe this is just the cost of integrating. Maybe you could run `dep prune` to see if we can reduce the size of `vendor/`.

Dockerfile
```diff
@@ -50,4 +50,4 @@ COPY --from=compress /gomplate-slim /gomplate
 
 ENTRYPOINT [ "/gomplate" ]
 
-CMD [ "--help" ]
\ No newline at end of file
+CMD [ "--help" ]
```
hairyhenderson: unrelated to this PR

test/integration/datasources_k8s.bats
```diff
@@ -0,0 +1,23 @@
...
+
+}
+
+function teardown () {
+  stop_k8s_svc
+  rm -rf /var/run/secrets/kubernetes.io
+}
+
+@test "Testing k8s" {
+  gomplate -i '{{k8s.ConfigMap "default" "test" "test.property.1" "default"}}'
```
hairyhenderson: This should look more like this:

```bash
gomplate -d k8s=k8s:// -i '{{(datasource "k8s" "default" "test" "test.property.1" "default" )}}' 
```

(though it probably makes sense for the syntax to be quite a bit less complex)

test/integration/datasources_k8s.bats
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bats
+
+load helper
+
+function setup () {
+  mkdir -p /var/run/secrets/kubernetes.io/serviceaccount
```
hairyhenderson: Would much rather a temporary directory be used here - these tests can be run outside of a container, and I don't want to risk accidentally clobbering something on someone's system.

test/integration/datasources_k8s.bats
```diff
@@ -0,0 +1,23 @@
...
+  touch /var/run/secrets/kubernetes.io/serviceaccount/token
+  cp /tmp/ca.crt  /var/run/secrets/kubernetes.io/serviceaccount/
```
hairyhenderson: This `/tmp/ca.crt` file won't exist when these tests are run outside of a container - would a relative path work?

test/integration/k8ssvc/ca.crt
```diff
@@ -0,0 +1,21 @@
+-----BEGIN CERTIFICATE-----
```
hairyhenderson: Even though these are only used for tests, I'd really rather not have key material stored in the repo. Can you re-jig the tests so that these certs are generated on-the-fly when necessary in `test/integration/datasources_k8s.bats`'s `startup` function?

Dockerfile
josledp: wop, don't know how this arrived to the PR

test/integration/datasources_k8s.bats
josledp: The point is that to get a configmap we need 3 parameters, the k8s namespace, to configmap name and the key we want to retrieve. About the  "-d k8s=k8s://" I can add it, but when inside a k8s the k8s client know how to connect, so there will be not any kind of uri there.

test/integration/datasources_k8s.bats
josledp: this is where the client is going to get the serviceaccount data needed to access de k8s api, it is not configurable, so I cannot do so much here. Anyway /var/run is a tmpfs (filesystem on memory), so if the teardown is not able to remove it, whenever you reboot these files will be gone

test/integration/datasources_k8s.bats
josledp: see the next comment about this. I will try...

test/integration/k8ssvc/ca.crt
josledp: Yeah, it is awful. I created the generate.sh and the openssl.cnf to make the cert generation automatically, sadly it is hard to setup in the integration test. Currently there are 3 differents ways of running integration tests ( the docker for CircleCi, the docker for local testing, and directly without docker). It would be much easier if there werer only 1 way of running integration tests.

test/integration/datasources_k8s.bats
josledp: The point is, the k8s configmap is more similar to the aws metadata, that to a datasource.

hairyhenderson:
> About integration test, I will then emulate the k8s API, because circle ci will not be able to spin up a minikube.

Would it make sense to instead use `localkube`? You could download it if not present during the test from https://github.com/kubernetes/minikube/releases/download/v0.24.1/localkube - I haven't used this, but according to the source:

> LocalkubeServer provides a fully functional Kubernetes cluster running entirely through goroutines

Seems like exactly what we need here

test/integration/datasources_k8s.bats
hairyhenderson: Hrm... Would a URL like `k8s:///namespace/configmap_name/key` make sense? This would be similar to the way the `vault://` URI scheme works.

You have a good point w.r.t. the `aws` functions. I did have a vague plan _at some point_ to make an AWS datasource, but that's more complicated because of the way their various APIs are arranged.

The advantage with using datasources is that it abstracts the source from the template. I could render the same template using a JSON file as a datasource in one scenario, or a Kubernetes ConfigMap in another scenario.

_For now_, let's please start with the datasource approach. If it becomes clear that a `k8s` namespace is also necessary, we can add it on afterwards.

test/integration/datasources_k8s.bats
hairyhenderson: Ah. `/var/run` may be `tempfs` on some OSes, but not macOS (where I sometimes run bare integration tests).

We'll leave this like this for now. Hopefully this path becomes configurable at some point...

test/integration/k8ssvc/ca.crt
hairyhenderson: The integration tests should run the same regardless of where they're running. If you call the cert generation script conditionally (i.e. skip if the certs exist already) from the `setup` function, it shouldn't be too big of an impact.

hairyhenderson:
> Seems like exactly what we need here

I take it back - `localkube` is not the full story... However, this might work with CircleCI's remote docker thingy: https://github.com/kubernetes/minikube/tree/master/deploy/docker

josledp:
I modified the k8s to be a datasource. I have set the URI as k8s://namespace/configmap_name, so you can access different keys on the datasource
I've left the emulated k8s for integration tests, running localkube would need to run gomplate inside a pod (docker) inside the docker for testing. I think this is not possible in CircleCI.

hairyhenderson:
Thanks @josledp - will take a look at that over the next little while.

I noticed you did a `dep prune` which removed a lot of other dependencies (it's been a while since I've pruned, clearly 😉). I've done a `dep prune` on `master` (in #237) - can you rebase to `master` to clear those changes out of this PR?

josledp:
rebased

hairyhenderson:
@josledp the CI test is failing:

```
not ok 27 Testing k8s
# (from function `setup' in test file test/integration/datasources_k8s.bats, line 8)
#   `cp /tmp/ca.crt  /var/run/secrets/kubernetes.io/serviceaccount/' failed
# cp: cannot stat '/tmp/ca.crt': No such file or directory
```

data/datasource.go
```diff
@@ -426,6 +431,36 @@ func readConsul(source *Source, args ...string) ([]byte, error) {
 	return data, nil
 }
 
+func readK8S(source *Source, args ...string) ([]byte, error) {
+	if len(args) < 1 {
+		return nil, fmt.Errorf("a key is needed")
```
hairyhenderson: Hmm... Why not allow a missing key, and return the whole `configmap.Data` object? Then you could do something like:

```
$ gomplate -d k=k8s:///default/map -i '{{ (ds "k").foo }}'
hello world
```

Where the configmap's data looks like:

```json
{ "foo": "hello world", "bar": "goodbye world" }
```

Obviously simply specifying `{{ds "k" "foo"}}` would have the same effect, but I think the flexibility is useful, especially when doing things like iterating though all of the keys in the map:

```
$ gomplate -d k=k8s:///default/map -i '{{ range $key, $value := (ds "k") }}{{ $key }} is {{ $value }}{{"\n"}}{{end}}'
bar is goodbye world
foo is hello world
```

data/datasource.go
```diff
@@ -426,6 +431,36 @@ func readConsul(source *Source, args ...string) ([]byte, error) {
...
+	configmapName := strings.Trim(source.URL.Path, "/")
+	configmap, err := source.K8S.CoreV1().ConfigMaps(namespace).Get(configmapName, meta_v1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	if data, ok := configmap.Data[args[0]]; ok {
+		return []byte(data), nil
+	}
+	if len(args) > 1 {
+		return []byte(args[1]), nil
```
hairyhenderson: The `datasource` function doesn't support default values, and I'd rather not add that support to this specific datasource - so please remove this block.

There was some discussion of a `default` function in #213 that would solve the general case.

funcs.go
```diff
@@ -21,5 +21,6 @@ func initFuncs(d *data.Data) template.FuncMap {
 	funcs.AddTimeFuncs(f)
 	funcs.AddMathFuncs(f)
 	funcs.AddCryptoFuncs(f)
+
```
hairyhenderson: Remove this extra line please

test/integration/Dockerfile
```diff
@@ -23,6 +24,10 @@ COPY gomplate /bin/gomplate
 COPY mirror /bin/mirror
 COPY meta /bin/meta
 COPY aws /bin/aws
+COPY k8s /bin/k8s
+COPY k8ssvc/generate.sh /tmp/
+COPY k8ssvc/openssl.cnf /tmp/
+RUN /tmp/generate.sh
```
hairyhenderson: Can the script be run from the test itself, at runtime, instead of build-time? If it's run at build-time the test will only work when the test is run inside a container.

hairyhenderson:
@josledp I think I'd like the TLS stuff to be removed, and support added to access the API over non-TLS. It looks like by default there is a non-TLS-protected `localhost:8080` API port listening (see https://kubernetes.io/docs/admin/accessing-the-api/#api-server-ports-and-ips). While it'd be useful to test out the TLS support, the setup mechanism is a bit too complex IMO and slows down the tests too much.

Maybe this behaviour could be controlled with a `KUBERNETES_TLS` environment variable (default to `true`, disable with `false`)?

data/datasource.go
josledp: How do I have to return the whole configmap.Data? Shoud a I return de []byte json?

josledp:
As far as I know, there is only a TLS endpoint available from a pod. I have not found an easy way to integrate the SSL part without embedding it in the image build process :/

data/datasource.go
hairyhenderson: Yeah, you'd just return it as a JSON-encoded `string` converted to a `[]byte`

hairyhenderson:
Sorry for the huge delay - was out of the country over Christmas and haven't fully caught up yet 😉

> As far as I know, there is only a TLS endpoint available from a pod.

I haven't had any luck either 😞, but I'll keep digging.

The CI build is still failing:

```
not ok 27 Testing k8s
# (from function `setup' in test file test/integration/datasources_k8s.bats, line 8)
#   `cp /tmp/ca.crt  /var/run/secrets/kubernetes.io/serviceaccount/' failed
# cp: cannot stat '/tmp/ca.crt': No such file or directory
```

Also there are a few other outstanding issues that need to be addressed. No rush though, and thank you for persisting on this!

hairyhenderson:
Almost a year later with no activity, I'm going to close this PR for now.

Now that I have far more experience with Kubernetes than I did when this was first suggested, support for ConfigMaps is really only relevant outside of pods. The usual pattern that I've seen is to mount the ConfigMap as a file inside the pod, and gomplate can read these files with no issue. I've seen gomplate used like this in a few cases.

If there is renewed interest, feel free to re-open the PR, though the updates necessary to rebase are going to be significant!

PR #251, opened 2018-02-04
Allow vault address to be specified in the vault:// URL

hairyhenderson:
Also updating the Vault & Consul clients used in the tests, and updating the tests for new syntax.


Makefile
```diff
@@ -104,7 +104,7 @@ test-integration-docker: build-integration-image
 	docker run -it --rm gomplate-test
 
 test-integration: build build-mirror build-meta build-aws
-	@test/integration/test.sh
+	@bats test/integration/
```
hairyhenderson: foo

PR #268, opened 2018-03-03
strings.Indent should not indent when width is 0

keitwb:
When recursively generating YAML, it is useful to have a simple generic template definition that starts with an indent of 0 and increments at each level YAML.  If you pass `width` of 0 to the function as it was before this change, it would indent it one level without complaining, which was a great source of confusion to me.


hairyhenderson:
Hi @keitwb! Thanks for taking the time to fix this. Not only is this confusing - it's a bug 😉

PR #271, opened 2018-03-04
Adding sockaddr namespace

### LINKED ISSUE #145: Add `go-sockaddr` functions

hairyhenderson:
Fixes #145 

Adds a pile of functions from the [`go-sockaddr`](https://github.com/hashicorp/go-sockaddr) package:

- `sockaddr.GetAllInterfaces`
- `sockaddr.GetDefaultInterfaces`
- `sockaddr.GetPrivateInterfaces`
- `sockaddr.GetPublicInterfaces`
- `sockaddr.Sort`
- `sockaddr.Exclude`
- `sockaddr.Include`
- `sockaddr.Attr`
- `sockaddr.Join`
- `sockaddr.Limit`
- `sockaddr.Offset`
- `sockaddr.Unique`
- `sockaddr.Math`
- `sockaddr.GetPrivateIP`
- `sockaddr.GetPrivateIPs`
- `sockaddr.GetPublicIP`
- `sockaddr.GetPublicIPs`
- `sockaddr.GetInterfaceIP`
- `sockaddr.GetInterfaceIPs`

I tried to closely match the function names that are exported in [github.com/hashicorp/go-sockaddr/template](https://github.com/hashicorp/go-sockaddr/tree/master/template), though couldn't directly import the `FuncMap`s defined therein, because I'd like these to be namespaced.

/cc @jen20 @sean-

### DIFF
--- a/funcs/sockaddr.go
+++ b/funcs/sockaddr.go
@@ -1,0 +1,122 @@
+package funcs
+
+import (
+	"sync"
+
+	"github.com/hashicorp/go-sockaddr"
+	"github.com/hashicorp/go-sockaddr/template"
+)
+
+var (
+	sockaddrNS     *SockaddrFuncs
+	sockaddrNSInit sync.Once
+)
+
+// SockaddrNS - the sockaddr namespace
+func SockaddrNS() *SockaddrFuncs {
+	sockaddrNSInit.Do(func() { sockaddrNS = &SockaddrFuncs{} })
+	return sockaddrNS
+}
+
+// AddSockaddrFuncs -
+func AddSockaddrFuncs(f map[string]interface{}) {
+	f["sockaddr"] = SockaddrNS
+}
+
+// SockaddrFuncs -
+type SockaddrFuncs struct{}
+
+// GetAllInterfaces -
+func (f *SockaddrFuncs) GetAllInterfaces() (sockaddr.IfAddrs, error) {
+	return sockaddr.GetAllInterfaces()
+}
+
+// GetDefaultInterfaces -
+func (f *SockaddrFuncs) GetDefaultInterfaces() (sockaddr.IfAddrs, error) {
+	return sockaddr.GetDefaultInterfaces()
+}
+
+// GetPrivateInterfaces -
+func (f *SockaddrFuncs) GetPrivateInterfaces() (sockaddr.IfAddrs, error) {
+	return sockaddr.GetPrivateInterfaces()
+}
+
+// GetPublicInterfaces -
+func (f *SockaddrFuncs) GetPublicInterfaces() (sockaddr.IfAddrs, error) {
+	return sockaddr.GetPublicInterfaces()
+}
+
+// Sort -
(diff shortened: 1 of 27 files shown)


funcs/sockaddr.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/hashicorp/go-sockaddr"
+	"github.com/hashicorp/go-sockaddr/template"
+)
+
+var (
+	sockaddrNS     *SockaddrFuncs
+	sockaddrNSInit sync.Once
+)
+
+// SockaddrNS - the crypto namespace
```
sean-: Cargo-culted `crypto`?

test/integration/sockaddr_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+var _ = Suite(&SockaddrSuite{})
+
+func (s *SockaddrSuite) TestSockaddr(c *C) {
+	result := icmd.RunCommand(GomplateBin, "-i",
+		`{{ range (sockaddr.GetAllInterfaces | sockaddr.Exclude "type" "ipv6") -}}
+{{ . | sockaddr.Attr "address" }}
+{{end}}`)
+	result.Assert(c, icmd.Expected{ExitCode: 0, Out: "127.0.0.1"})
+}
```
sean-: I need to go back and write a helper function that's builtin to `go-sockaddr` that makes testing easier.

hairyhenderson: the ability to inject some fake interface/address state to `go-sockaddr` would be useful for testing, but not entirely crucial...

PR #281, opened 2018-04-05
New function proposal: `files.Walk`

christopher-avila:


hairyhenderson:
Hi @christopher-avila, thank you for this contribution!

Can you please describe what you're aiming for with `file.FindFiles`? You should be able to accomplish the same effect with a template similar to this:

```console
$ gomplate -i '{{ range file.ReadDir "." }}{{ if not (file.IsDir .) }}{{ . }} IS A FILE{{"\n"}}{{ end }}{{ end -}}'
Gopkg.toml IS A FILE
LICENSE IS A FILE
CHANGELOG.md IS A FILE
...
```

As for `strings.TrimPrefix` - that seems like a reasonable addition, but I'd rather it be adapted to work better in a pipeline. This means that I should be able to use it in a template like `{{ print "Hello, World!" | strings.TrimPrefix "Hello, " }}`. To do this the signature needs its arguments swapped so the input string is last. See the `strings.Trim` function -  https://github.com/hairyhenderson/gomplate/pull/281/files#diff-3011984e758821c934ac913b336f80d3L82 - as an example.

Ideally I like to also have tests and documentation added for any new functions. If you're not comfortable with these I can add them in after.

Thanks!

robermorales:
Any update on this?

christopher-avila:
I'm modifying the `files.FindFiles` function, naming it `files.Walk`. It looks like a better name for the function (is actually a wrapper over `filepath.Walk`). I'll commit this changes on this PR and will open another one for `strings.TrimPreffix`.

Thanks for your time!

christopher-avila:
From: https://ci.appveyor.com/project/hairyhenderson/gomplate/build/637

```
--- FAIL: TestFileWalk (0.00s)
	Error Trace:	file_test.go:47
	Error:      	Not equal: 
	            	expected: []string{"/tmp", "/tmp/bar", "/tmp/bar/baz"}
	            	actual  : []string{"/tmp", "\\tmp\\bar", "\\tmp\\bar\\baz"}
```

Any idea of this? =/

hairyhenderson:
Thanks @christopher-avila!

I'm thinking the issue has to do with Windows path separators. If you instead use `filepath.Join` to create the expected paths you should _in theory_ be OK. I'm not sure why the first element is `/tmp` though... Perhaps an odd quirk in how afero's `Walk` is implemented.

christopher-avila:
I think it is solved. Take a look! 

Thanks for your time! :)

hairyhenderson:
Thanks again for the contribution, @christopher-avila!

christopher-avila:
Thanks to you for your time!

PR #282, opened 2018-04-12
New function proposal: `strings.TrimPrefix`

christopher-avila:
Test also added.


hairyhenderson:
Note to self: this still needs to be documented 😉

PR #287, opened 2018-04-13
meta: Add release for freebsd-amd64

jen20:
This pull request enables release builds for FreeBSD on amd64 - my test shows that all is well building, so I don't think any other modifications need to be made!


hairyhenderson:
Thanks @jen20! 🙂

PR #297, opened 2018-04-19
Preserve FileMode of input file when writing output file

djgilcrease:
Added the current files modes to be tracked though the gatherTemplates process, ensuring the current default of 644 if the input file is a string

Added tests for openOutFile and the full gatherTemplates

### DIFF
--- a/gomplate.go
+++ b/gomplate.go
@@ -2,6 +2,7 @@ package gomplate
 
 import (
 	"io"
+	"os"
 	"text/template"
 
 	"github.com/hairyhenderson/gomplate/data"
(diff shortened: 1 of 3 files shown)


djgilcrease:
looking into the panic and why manually running the exact same command does not panic locally for me

hairyhenderson:
This is looking like a good start!

Instead of adding a new type, I wonder if it'd be cleaner if `walkDir` could output a `[]*tplate` instead of the `[]string` for the input files - and then the `tplate` struct could be modified to include a `mode os.FileMode` field.

Does that sound OK to you?

djgilcrease:
Ya, I was thinking that at first. Will go redo the change.

PR #301, opened 2018-04-20
Preserve FileMode of input file when writing output file

### LINKED ISSUE #296: File mode is not preserved
when using something like ```gomplate --input-dir=templates --output-dir=deploy``` that contains files with the execute bit set that is not preserved.

I am thinking to update ```gatherTemplates``` to add get the current FileMode and pass it to ```addTarget``` and then into ```openOutFile```

If this approach is ok I will create the pull request

djgilcrease:
add mode to the tplate struct and update code to make use of this.

Fixes #296

### DIFF
--- a/template.go
+++ b/template.go
@@ -20,9 +20,11 @@ var Stdout io.WriteCloser = os.Stdout
 
 // tplate - models a gomplate template file...
 type tplate struct {
-	name     string
-	target   io.Writer
-	contents string
+	name       string
+	targetPath string
+	target     io.Writer
+	contents   string
+	mode       os.FileMode
 }
 
 func (t *tplate) toGoTemplate(g *gomplate) (*template.Template, error) {
@@ -41,9 +43,12 @@ func (t *tplate) loadContents() (err error) {
 	return err
 }
 
-func (t *tplate) addTarget(outFile string) (err error) {
+func (t *tplate) addTarget() (err error) {
+	if t.name == "<arg>" && t.targetPath == "" {
+		t.targetPath = "-"
+	}
 	if t.target == nil {
-		t.target, err = openOutFile(outFile)
+		t.target, err = openOutFile(t.targetPath, t.mode)
 	}
 	return err
 }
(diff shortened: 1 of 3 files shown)


djgilcrease:
I noticed that the linting does not check any of the sub packages, is this intentional?

hairyhenderson:
> I noticed that the linting does not check any of the sub packages, is this intentional?

Ack! No! Thanks for pointing that out 😀 - I'll fix that very soon (along with all the new-found linting errors 🤦‍♂️)

template.go
```diff
@@ -93,31 +103,29 @@ func gatherTemplates(o *Config) (templates []*tplate, err error) {
 // walkDir - given an input dir `dir` and an output dir `outDir`, and a list
 // of exclude globs (if any), walk the input directory and create a list of
 // input and output files, and an error, if any.
```
hairyhenderson: please update this comment now that the signature's changed 🙂

vault/auth.go
```diff
@@ -198,7 +198,7 @@ func (v *Vault) EC2Login() string {
 			nonce = val
 		}
 		fs := vfs.OS()
-		f, err := fs.OpenFile(output, os.O_WRONLY | os.O_CREATE | os.O_TRUNC, os.FileMode(0600))
+		f, err := fs.OpenFile(output, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, os.FileMode(0600))
```
hairyhenderson: any reason this change is part of this PR?

hairyhenderson:
@djgilcrease this looks decent. I see an empty file `tmp/test1` in the commit that probably shouldn't be there, and there's a few minor nitpicks, but otherwise I think it's good to go (after rebasing).

I've never really been fully satisfied with how the template gathering works, and I _think_ this moves things in the right direction. Hopefully some day I'll have a chance to sit down and refactor things 😉

vault/auth.go
djgilcrease: Gofmt change

hairyhenderson: ok, if you rebase I think this'll go away - this was done in #302

hairyhenderson:
Thanks for the contribution @djgilcrease!

PR #307, opened 2018-04-24
Execute additional command after --

### LINKED ISSUE #300: Exec mode...
One of the things that keeps coming up is the use of `gomplate` as a pre-processor followed by the execution of a binary.  It would be nice if it were possible to do something like:

```
gomplate ${GOMPLATE_FLAGS} --exec="consul -config-dir /usr/local/etc/consul.d/"

# or a more preferred option:

gomplate ${GOMPLATE_FLAGS} -- consul -config-dir /usr/local/etc/consul.d/
```

so that `gomplate` can literally be used as a preprocessor without resorting to a shell script to stitch these two commands together.

#featurerequest

hairyhenderson:
Fixes #300 

### DIFF
--- a/cmd/gomplate/main.go
+++ b/cmd/gomplate/main.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 
 	"github.com/hairyhenderson/gomplate"
 	"github.com/hairyhenderson/gomplate/env"
@@ -45,6 +46,29 @@ func printVersion(name string) {
 	fmt.Printf("%s version %s\n", name, version.Version)
 }
 
+// postRunExec - if templating succeeds, the command following a '--' will be executed
+func postRunExec(cmd *cobra.Command, args []string) error {
+	if len(args) > 0 {
+		name := args[0]
+		args = args[1:]
+		c := exec.Command(name, args...)
+		c.Stdin = os.Stdin
+		c.Stderr = os.Stderr
+		c.Stdout = os.Stdout
+		return c.Run()
+	}
+	return nil
+}
+
+// optionalExecArgs - implements cobra.PositionalArgs. Allows extra args following
+// a '--', but not otherwise.
+func optionalExecArgs(cmd *cobra.Command, args []string) error {
+	if cmd.ArgsLenAtDash() == 0 {
+		return nil
+	}
+	return cobra.NoArgs(cmd, args)
+}
+
 func newGomplateCmd() *cobra.Command {
 	rootCmd := &cobra.Command{
 		Use:     "gomplate",
@@ -57,7 +81,8 @@ func newGomplateCmd() *cobra.Command {
 			}
 			return gomplate.RunTemplates(&opts)
 		},
-		Args: cobra.NoArgs,
+		PostRunE: postRunExec,
+		Args:     optionalExecArgs,
 	}
 	return rootCmd
 }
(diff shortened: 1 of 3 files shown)


cmd/gomplate/main.go
```diff
@@ -45,6 +46,29 @@ func printVersion(name string) {
 	fmt.Printf("%s version %s\n", name, version.Version)
 }
 
+// postRunExec - if templating succeeds, the command following a '--' will be executed
```
osterman: It might be good to not "overload" the `--` to only mean execution. That way it can be used in other contexts down the line. 

The way `chamber` and `aws-vault` work is they have an additional argument to indicate execution. 

https://github.com/segmentio/chamber#exec
https://github.com/99designs/aws-vault#usage

hairyhenderson: 🤔 do you have other contexts in mind?

I had a brief look at those other tools and it seems they both have an explicit `exec` subcommand.

We _could_ do that in `gomplate`, but I'm not sure I really see the point right yet. At this point it seems like it'd just be an extra 4 characters to type 😉

There's nothing stopping alternate usage with a subcommand later on, while still supporting this more basic syntax.

Does that make sense @osterman?

osterman: Yes - makes sense. 

I look at it in terms of future proofing the command line interface. While today, it's adequate, it might reduce confusion in the long run. The way `gomplate` is implemented is not unconventional or unreasonable.  =) Just have noticed that amongst the apps I use these days, a common interface has started to evolve. If you look at `helm`, `kops`, `kubectl`, `chamber`, `aws-valut`, `terraform`,  `vault`, `tctl` they all implement them this way. (They all happen to be in `go` as well)

This is quickly spiraling into a larger conversation and just so you know, I don't feel incredibly strongly about it. I just wanted to point this out, since I think it's a nice interface to copy. 

In `gomplate`, the `--version` option is available. There's no disputing that this is *incredibly* common of course in console apps. However, the trend I'm seeing to leverage a combination of positional parameter aka subcommand with options. I guess the thinking is options modify commands. So `--version` doesn't modify anything. It's a command. This will make it easier as well to scope options to specific commands. 

In the case of `gomplate`, I could envision a number of subcommands. 

e.g
`validate`
`exec`
`version`
`apply`
`help`

hairyhenderson: Good point @osterman, and, I'm on board. Maybe at some point later on I'll add some of those. I like the idea of `validate` too, though not _entirely_ certain that'd be easy to implement.

Anyway, for now we'll stick with this current naive/simplistic approach I think 😉

sean-: This isn't a bad idea, actually.

```
gomplate exec ${normal_gomplate_args} -- /my/command/and/its/args
```

Whereas the default subcommand is `gomplate render` ?  Then it would be clear and easy to say, "the `exec` subcommand requires the `--` delimiter to separate `gomplate` arguments from the command to be executed by `gomplate`".  If you don't specify `exec` or `render`, then you're getting the author's best guess at defaults and intentions.

Regardless, I'm excited for this.

osterman:
Thanks @hairyhenderson !

osterman:
Btw, will you be cutting a new release that includes this change? Would like to download the binary. =)

hairyhenderson:
@osterman I will - I think 2.5.0 is long overdue, so I'll be releasing that as soon as I get some time 😉 - hopefully tonight (EDT)

osterman:
We're using this here: https://github.com/cloudposse/default-backend/blob/master/Dockerfile#L11

Thanks!

hairyhenderson:
that's awesome, @osterman - I'm glad it's useful for you! 🎉

Issue #328, opened 2018-05-14
The `sockaddr.Include` and `sockaddr.Exclude` do not have "private" selector.


kwilczynski:
Hi @hairyhenderson,

I've noticed that the documentation mentions that the `sockaddr.Exclude` and `sockaddr.Include` support a "private" selector, but it does not seem to be the case, for instance:

```
$ gomplate -i '{{ sockaddr.GetAllInterfaces | sockaddr.Include "private" "" }}'
Error: template: <arg>:1:39: executing "<arg>" at <sockaddr.Include>: error calling Include: invalid include selector "private"
```

Which is correct as per the [ifaddrs.go](https://github.com/hashicorp/go-sockaddr/blob/6d291a969b86c4b633730bfc6b8b9d64c3aafed9/ifaddrs.go#L963-L994) implementation.

To achieve this (somewhat), one would have to use the `rfc` selector with an appropriate numeric RFC code. Alternatively, we could send a Pull Request against `go-sockaddr` to add this include/exclude filter.

Would you like to send a Pull Request to update documentation?

kwilczynski:
I have opened a Pull Request at https://github.com/hashicorp/go-sockaddr/pull/23 to add support for `private` to the `include` and `exclude` filters.

hairyhenderson:
Hi @kwilczynski! Thanks for logging this issue. I'm really not sure why I documented `private` in there - probably a copy/paste error from `Sort`.

If you'd like to issue a PR to fix the doc, that'd be great!

From what I can see, doing `sockaddr.Include "rfc" "6890"` should accomplish what you're looking for. With `Sort`, `private` is really just a wrapper around the RFC 6890 comparator.

kwilczynski:
Hi @hairyhenderson,

No worries! Fix to the documentation coming right up!

hairyhenderson:
Fixed in #329

PR #329, opened 2018-05-17
Remove notion of "private" selector from the Include/Exclude filter.

kwilczynski:
This commit removes any mention of the "private" selector from the Include and
Exclude filters, as it is not supported by the upstream `go-sockaddr` project.

Signed-off-by: Krzysztof Wilczynski <kw@linux.com>

### DIFF
--- a/docs/content/functions/sockaddr.md
+++ b/docs/content/functions/sockaddr.md
@@ -84,7 +84,6 @@ The valid selectors are:
 | `default` | whether or not the `IfAddr` has a default route |
 | `name` | the interface name |
 | `port` | the port, if included in the `IfAddr` |
-| `private` | private addresses first |
 | `size` | the size of the network mask, smaller mask (larger number of hosts per network) to largest (e.g. a /24 sorts before a /32) |
 | `type` | the type of the `IfAddr`. Order is Unix, IPv4, then IPv6 |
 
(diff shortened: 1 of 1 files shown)


kwilczynski:
Fixes #328.

docs/content/functions/sockaddr.md
```diff
@@ -372,4 +370,4 @@ $ gomplate -i '{{ sockaddr.GetInterfaceIPs "en0" }}'
 10.0.0.28 fe80::1f9a:5582:4b41:bd18
 ```
 
-[RFC 6890]: http://tools.ietf.org/html/rfc6890
+[1]: http://tools.ietf.org/html/rfc6890
```
hairyhenderson: why this change?

hairyhenderson: The `[RFC 6890][]` links are totally valid markdown - they all produce correct HTML:

```html
<a href="http://tools.ietf.org/html/rfc6890">RFC 6890</a>
```

There's an `[RFC 1918][]` link, on the other hand, which is invalid, because there's no associated `[RFC 1918]: http://tools.ietf.org/html/rfc1918` footer (which is a separate bug).

Since this PR is about removing `private`, please revert the `RFC 6890` changes.

kwilczynski: Ah, that's my bad. I meant the broken one. Apologies.

kwilczynski:
Hi @hairyhenderson,

TL;DR
I removed the wrong fix to the markdown label (will send other Pull Request to resolve that) and only pushed change to remove the mention of "private". 

There are some lessons learned here for me :)

1) The economy class on AirFrance does not have enough space for 15" MacBook Pro
2) The "tap to click" and turbulences do not complete each-other well :) Doh.
3) Paying $10 (or so) for 200 MB of data was silly (albeit, I can't sleep doing the flights so talked myself into trying to be more "productive", so to speak)

Re: closing the Pull Request - for some cosmic reason I managed to tap the trackpad by when plane was flying through a small area of turbulence when the cursor was right above "Close pull request" when I was about to write a response. It kind of made me laugh a bit back then :) Apologies for troubles.

PR #331, opened 2018-05-18
Adding more math functions

hairyhenderson:
More functions in the `math` namespace:

- `math.Abs`
- `math.IsInt`
- `math.IsFloat`
- `math.IsNum`
- `math.Max`
- `math.Min`
- `math.Ceil`
- `math.Floor`
- `math.Round`

Most other functions now also support floating-point arithmetic if at least one of the inputs is a float.

Plus, all functions return an `interface{}` instead of an explicit `int64` (or `float64`).

### DIFF
--- a/docs/content/functions/math.md
+++ b/docs/content/functions/math.md
@@ -9,13 +9,7 @@ A set of basic math functions to be able to perform simple arithmetic operations
 
 ### Supported input
 
-_**Note:** currently, `gomplate` supports only integer arithmetic. All functions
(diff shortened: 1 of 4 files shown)


kwilczynski:
Hi @hairyhenderson, 

This is amazing! Do you think "abs" would be also useful?

hairyhenderson:
> Do you think "abs" would be also useful?

I don't see why not! I'll toss it in while I'm at it 😉

Issue #370, opened 2018-08-15
Overriding MIME Types doesn't work


vadim-hleif:
### Works
 type of content  the same as in `Content-Type` response header:
* `gomplate -d config=http://private-30833-vadimhleif.apiary-mock.com/yaml  -i '{{ (ds "config").name }}'`
### Don't work
 type of content  not the same as in `Content-Type` response header, overriding MIME doesn't work:
* `gomplate -d config=http://private-30833-vadimhleif.apiary-mock.com/text   -i '{{ (ds "config").name }}'` don't work, expected
* `gomplate -d config=http://private-30833-vadimhleif.apiary-mock.com/text?type=application/yaml   -dicofnig=s "config").name }}'` don't work, not expected

![capture](https://user-images.githubusercontent.com/11632146/44151705-07157cee-a0ac-11e8-92f8-c894e37558a2.PNG)

### what
It may be helpful. for example, for github api: 
* `https://raw.githubusercontent.com/cloudposse/geodesic/master/README.yaml`
### why
* Github return resource with `text/plain` content-Type and gomplate can't parse yaml

### references
- https://gomplate.hairyhenderson.ca/datasources/#mime-types
- https://github.com/cloudposse/build-harness/issues/87

hairyhenderson:
Hi @vadim-hleif, thanks for reporting this! I can confirm this behaviour, and based on how I wrote it, it makes sense - the URL query string gets parsed long before the `Content-Type` header is seen, so in effect, the `Content-Type` header always overrides anything provided in the URL.

I'd tend to agree with you that it's a bug, since the express purpose of the `?type=application/yaml` query string is to force the type.

My initial feeling on precedence is: path extension, `Content-Type` header, then finally, `type` query. 

This wouldn't help you in the case of the GitHub URL provided, since the `Content-Type` would still take precedence over the path extension. That being said, the `raw.githubusercontent.com` host will only return `text/plain` for non-binary files, which I seem to recall is by design (but can't find the reference right now)...

An option in the GitHub case is to use their contents API. It's a bit of a kludge but it's more predictable:

```
$ gomplate -d gh=https://api.github.com/repos/cloudposse/geodesic/contents/README.yaml -i '{{ ((ds "gh").content | base64.Decode | data.YAML).name }}'
Geodesic
```

Either way, `?type` should have precedence over `Content-Type`, so I'll fix that.

vadim-hleif:
[...]

PR #373, opened 2018-08-17
Saving type from get param

vadim-hleif:
### what
* Add flag for checking - is `type` param passed or not

### why
* Previously type of resource always gets from response header `Content-Type` and override by get param `type` was impossible 

### references
- https://github.com/hairyhenderson/gomplate/issues/370

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ report.xml
 *.cid
 *.iid
 *.out
+.vscode/*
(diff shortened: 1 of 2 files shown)


hairyhenderson:
Thanks for submitting this, @vadim-hleif!

I've actually been working on a different solution, but I've been offline (travelling back from vacation) so haven't had the chance to issue the PR yet. I hope you don't mind, but I'd rather use the changes I've been working on instead of this.

hairyhenderson:
see #374

PR #392, opened 2018-09-07
[WIP] Add test.Required function

### LINKED ISSUE #310: Support for `required`
## what
A `required` function that allows you to declare a particular values entry as required for template rendering. If the value is empty, the template rendering will fail with a user submitted error message.

This function is available in `helm`. 

## why
* Sometimes we need to ensure user has defined specific environment variables or the rendered template is invalid. 
* Outputting helpful messages will make this easier

## example
```
{{- required "A valid API_KEY entry required!" .Env.API_KEY -}}
```

## references:
- https://github.com/kubernetes/helm/blob/master/docs/charts_tips_and_tricks.md#know-your-template-functions

hairyhenderson:
Fixes #310 


docs-src/content/functions/test.yml
```diff
@@ -37,3 +37,43 @@ funcs:
         template: <arg>:1:3: executing "<arg>" at <fail>: error calling fail: template generation failed
         $ gomplate -i '{{ test.Fail "something is wrong!" }}'
         template: <arg>:1:7: executing "<arg>" at <test.Fail>: error calling Fail: template generation failed: something is wrong!
+  - name: test.Required
+    alias: required
+    description: |
+      Passes through the given value, if it's non-empty, non-zero, and non-`nil`.
```
hairyhenderson: Hmm... I may want to re-think this. Helm only cares if it's an empty string or `nil`, whereas my implementation looks at _any_ zero value, including `0`, `false`, empty arrays, and empty maps. This may be somewhat less useful. And besides, if the goal is to prevent specific values (like `0`), `assert` is probably more appropriate.

osterman: Consider this:

```
nodes: {{ getenv "NODE_COUNT" | required "Must specify NODE_COUNT; use `0` for none" }}
```

I think `0` is a non-error. But assert could also be okay, I suppose. Anyways, as far as *our* use-case goes, empty (as in not defined) is what we care about.

docs/content/functions/test.md
```diff
@@ -69,3 +69,57 @@ template: <arg>:1:3: executing "<arg>" at <fail>: error calling fail: template g
...
+```
+```console
+$ cat <<EOF> config.yaml
+defined: a value
+empty: ""
+EOF
+$ gomplate -d config=config.yaml -i '{{ (ds "config").defined | required "The `config` datasource must have a value defined for `defined`" }}'
+a value
+$ gomplate -d config=config.yaml -i '{{ (ds "config").empty | required "The `config` datasource must have a value defined for `empty`" }}'
+template: <arg>:1:25: executing "<arg>" at <required "The `confi...>: error calling required: The `config` datasource must have a value defined for `empty`
```
osterman: perfect!

docs/content/functions/test.md
```diff
@@ -69,3 +69,57 @@ template: <arg>:1:3: executing "<arg>" at <fail>: error calling fail: template g
...
+$ gomplate -d config=config.yaml -i '{{ (ds "config").bogus | required "The `config` datasource must have a value defined for `bogus`" }}'
+template: <arg>:1:7: executing "<arg>" at <"config">: map has no entry for key "bogus"
```
osterman: Is there a way we could do something like:
```
{{ (ds "config") | get "bogus" | required "The `config` datasource must have a value defined for `bogus`"
```

where `get` works like `getenv` and does not error?

osterman: haven't tried...

docs-src/content/functions/test.yml
hairyhenderson: @osterman yeah, undefined is unfortunately not going to be handled by `required` (
https://github.com/hairyhenderson/gomplate/issues/310#issuecomment-411835644), because gomplate uses the `missingkey=error` option (and Helm doesn't).

But I think it's probably fair to scope this down to only failing on `nil` and `""` (empty string), which is what Helm does for defined values.

docs/content/functions/test.md
hairyhenderson: ah, that might work. There is the `index` built-in, but it doesn't pipeline so well:

```console
$  bin/gomplate -i '{{ $j := (json `{"foo":"bar"}`)}}{{ index $j "baz" | required "nope!" }}'
template: <arg>:1:53: executing "<arg>" at <required "nope!">: error calling required: nope!
```

_if only_ `index` took arguments the other way around, it'd be a simple case of `{{ $j | index "baz" | required "nope!" }}`.

I could write an alternate for `index` that takes arguments in the opposite order for easier pipelining, but it may not be _that_ necessary...

docs-src/content/functions/test.yml
hairyhenderson: Updated to only fail on `nil` and `""`

PR #397, opened 2018-09-12
 Add Plugin proof of concept

mparadiso94:


hairyhenderson:
Hi @mparadiso94, thanks for the PR! This looks interesting, and I think I _mostly_ understand what's going on, but can you explain a little bit what your use-case with plugins is, and how this is intended to work?

Also, is `CGO_ENABLED=1` required?

From my understanding, plugins only work on Linux and macOS, which would make things difficult for the Windows, FreeBSD, and Solaris builds.

PR #400, opened 2018-09-13
adding a plugin proof of concept

mparadiso94:


hairyhenderson:
Looks like this is the same as #397 - as I mentioned there:

Hi @mparadiso94, thanks for the PR! This looks interesting, and I think I mostly understand what's going on, but can you explain a little bit what your use-case with plugins is, and how this is intended to work?

Also, is `CGO_ENABLED=1` required?

From my understanding, plugins only work on Linux and macOS, which would make things difficult for the Windows, FreeBSD, and Solaris builds.

hairyhenderson:
Closing for now, but feel free to reopen if you can think of a way to make this work with all platforms that gomplate supports.

Issue #210, opened 2017-09-24
nested template technique?


f4tq:
Is there a way to do this?
file data: 
```
---
a: 1
```
template1 data:
```
Welcome to {{.a}
{{ if .a == 1}} 
{{ template "foo" }}
{{ end - }}
```
It seems like `--input-dir` just renders everything unconditionally.  
It seems like `-d foo=template://foo.t` could be used treat conditional (nested) templates as datasources.  Since go templates need names, it fits with the `-d` scheme.  

I ran into trouble with `-d` and the use of a `.t` suffix trying to sniff mime typing so maybe a different switch with nearly identical behavior as `-d` that can be used for `define`ing templates for use in `{{template "foo"}}`...

Is there another way to do it?

Thanks for doing this!

hairyhenderson:
Hi @f4tq - thanks for logging this!

> It seems like `--input-dir` just renders everything unconditionally.

Right - it's intended for use with a directory containing _just_ templates. If you don't want to attempt to render certain files, it's probably best to wrap `gomplate` with a script which can differentiate between the types of files you want to render and those you don't.

> It seems like `-d foo=template://foo.t` could be used treat conditional (nested) templates as datasources. Since go templates need names, it fits with the `-d` scheme.

Interesting! It _does_ seem like it'd be a close fit. Right now, `--datasource`/`-d` is intended for use specifically with the `datasource`/`include` functions, and not `template` (which is a built-in). But I think for this use-case it's probably worth adding that support.

> I ran into trouble with `-d` and the use of a `.t` suffix trying to sniff mime typing ...

What sort of trouble? You _should_ be able to use an arbitrary file as input, then you can use [`include`](https://gomplate.hairyhenderson.ca/functions/data/#include) to include the data as-is. But I'm not sure this would work to define a template.

f4tq:
I [fixed this and more on my fork here](https://github.com/f4tq/gomplate/commit/9b12c6a45b10cc3793ace1c4ec134ba10a6a5418)
### Briefly
- I added `dict` and `slice` functions
- I removed the need for `(datasource "xyz").result.data` constructs but I left them in place
> I did not remove the cache concept nor the `datasource` function but they are un-necessary. I replaces the empty `Context` struct passed to template execution with resolved namespace passed from the cli with actual trees.
- I implemented nested templates

I am under the gun so I pressed on.  If you're willing, I'll submit a PR. 

### Overview
Given a data(source) like:
```data.yaml:
---
[...]

PR #403, opened 2018-09-18
new arg --templates for parsing additional templates

laher:
Does this seem like a good approach for loading additional templates like `{{ template "otherfile.tpl" }}`?

I think couldn't think what to call the CLI arg - maybe `--additional-templates` is better, but here's a straw man.

Cheers, great project, just this one thing missing IMO :+1: :100: :heart:

### DIFF
--- a/gomplate.go
+++ b/gomplate.go
@@ -1,8 +1,8 @@
 package gomplate
 
 import (
-	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -113,7 +113,7 @@ func (g *gomplate) runTemplate(t *tplate) error {
 			defer t.target.(io.Closer).Close()
 		}
 	}
-	err = tmpl.Execute(t.target, context)
+	err = tmpl.ExecuteTemplate(t.target, t.name, context)
 	return err
 }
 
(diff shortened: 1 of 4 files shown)


hairyhenderson:
Hi @laher, thanks for the PR!

I definitely want to implement support for nested templates, and there's already some discussion in #210 about the right way to implement that support.

So far the approach I'm leaning towards is this one: https://github.com/hairyhenderson/gomplate/issues/210#issuecomment-417150954

Would you mind commenting on that issue, and letting me know whether that approach would work for you, or if you have reasons that it wouldn't?

laher:
Interesting, thanks. I don't get why you'd treat templates as data, they're
exactly not the data. I'll reply later today in that thread with a more
thoughtful response :)

laher:
@hairyhenderson as discussed, I have implemented those local-filesystem use cases 1-4.

I did do something differently, but I can change it if it doesn't work for you: I found that the standard library `[template.ParseFiles(...)](https://golang.org/pkg/text/template/#Template.ParseFiles)` uses the basename of the file as the template name, so it felt more natural to follow suit with that.

i.e. instead of this: `gomplate --template path/to/foo.t -i '{{ template "path/to/foo.t" }}'`
it does this: `gomplate --template path/to/foo.t -i '{{ template "foo.t" }}'`

and instead of: `gomplate --template path/to/ -i '{{template "path/to/foo.t" }}'`
it does: `gomplate --template path/to/ -i '{{template "foo.t" }}'`

I did do the directory-aliasing prefix though: `gomplate --template bar=path/to/ -i '{{template "bar/foo.t" }}'`, and I took out the globbing.

What do you think?

hairyhenderson:
> I found that the standard library [template.ParseFiles(...)](https://golang.org/pkg/text/template/#Template.ParseFiles) uses the basename of the file as the template name, so it felt more natural to follow suit with that.

Yeah - I was a bit conflicted about that. The advantage is, of course, a  shorter template name, but the disadvantage is that templates with the same filename in different directories will conflict. It's a relatively minority edge-case, but I'd rather be more explicit about the name, especially when you consider that you can always use an alias like `--template foo.t=path/to/foo.t` to achieve the same effect.

Keeping the path as-is is also more in-line with what you get when you reference a directory, so I think I'd rather the internal consistency over the `template.ParseFiles` behaviour.

template.go
```diff
@@ -26,14 +26,30 @@ type tplate struct {
...
 	tmpl.Option("missingkey=error")
 	tmpl.Funcs(g.funcMap)
 	tmpl.Delims(g.leftDelim, g.rightDelim)
-	return tmpl.Parse(t.contents)
+	tp, err := tmpl.Parse(t.contents)
+	if err != nil {
+		return tp, err
+	}
+	for alias, path := range t.additionalTemplates {
+		b, err := ioutil.ReadFile(path)
```
hairyhenderson: you'll need to add `// nolint: gosec` above this line to get the linter to shut up:

```
template.go:43::warning: Potential file inclusion via variable,MEDIUM,HIGH (gosec)
```

It's more than a _potential_ file inclusion, it's a _deliberate_ file inclusion, but that's the whole point of this feature 😂

gomplate.go
```diff
@@ -108,15 +117,60 @@ func (g *gomplate) runTemplate(t *tplate) error {
...
-		rightDelim: rightDelim,
-		funcMap:    Funcs(d),
+		leftDelim:       leftDelim,
+		rightDelim:      rightDelim,
+		funcMap:         Funcs(d),
+		templateAliases: ta,
 	}
 }
 
+func parseTemplateArgs(templateArgs []string) (templateAliases, error) {
```
hairyhenderson: the linter's annoyed about this function having high cyclomatic complexity:

```
gomplate.go:132::warning: cyclomatic complexity 11 of function parseTemplateArgs() is high (> 10) (gocyclo)
```

If you see a simple way to refactor this to lower the complexity, that'd be great, otherwise, just add `// nolint: gocyclo` above this line and I'll maybe get to simplifying this _later_... 😉

hairyhenderson:
@laher Further to the comment above, this looks good so far, thank you for working on this!

And when you get the CI build to pass (should just be a case of appeasing the linter - see inline comments about that), I'll get this merged.

I'll follow up later with unit & integration tests, unless you want to add some yourself.

This'll also need some changes to documentation - but again, I can take care of that unless you feel like it 😉. It can certainly happen in another PR.

laher:
I got the CI passing, simplified `parseTemplateArgs`, and fixed some bugs in the process (mainly regarding slashes, paths & aliases)

Re testing: I've been testing manually on the CLI, it seems to work OK, but I think I'd need to rework the code in order to support the afero fs as in `TestGatherTemplates`. Not sure the best way to approach it at this point, and I'm not gonna get back to this for a few days now... Cheers

hairyhenderson:
Thanks @laher - this is great, and I'll merge it at this point. I totally understand needing to switch contexts for a few days, you've already been very generous with your time.

I'll merge this now, and will add some tests and docs for this.

PR #415, opened 2018-10-16
Remove CMD --help -> make call without args easy

strokyl:
Before this fix, it was really not easy to call gomplate without args.
For example if someone wanted to do:

> echo "Hello, {{.Env.USER}}" | docker run -e USER -i hairyhenderson/gomplate

This did not work, so he add to do:

> echo "Hello, {{.Env.USER}}" | docker run -e USER --entrypoint /gomplate -i hairyhenderson/gomplate


hairyhenderson:
Hi @strokyl - thanks for this contribution! Sorry I missed it!

I'm not sure why the CI build is failing - will try to kick it 😉

Issue #404, opened 2018-09-24
Add datasources to root context


hairyhenderson:
This is an _extraction_ from #210, where one of the items that @f4tq worked on was the removal of the need for the `datasource`/`ds` function, by adding datasources to the root context (i.e. the default `.` in a template).

I'm not 100% certain this can fully work, though the initial work done by @f4tq in [his fork](
https://github.com/f4tq/gomplate/commit/c4145ea086da1deeca0cecd278c63e0dad92d2b3) works for the simple case (i.e. datasource with no arguments).

This would make the following commands equivalent:

```console
$ gomplate -d config=some:///config/source -i '{{ (ds "config").value }}'
$ gomplate -d config=some:///config/source -i '{{ .config.value }}'
```

Still to figure out:
- what to do about datasources that take multiple (required or optional) arguments? do we ignore these and only support the simple case?
- a key designs around datasources is that they're lazily-loaded, and this needs to be retained - i.e. a datasource is only read when needed (i.e. when the `datasource` function is executed), and if `datasource` is never called (due to a negative `if` condition, for example), gomplate must not attempt to read the datasource at all
- we can't break current usage of `.Env` - this should be straightforward, though

ArcticSnowman:
Maybe take a look at Helm and how its' template processor includes the data as a `.Values` in the context.

hairyhenderson:
@ArcticSnowman helm doesn't load `.Values` dynamically (or, any of the other items in the context), so it doesn't have this challenge 😉.

I'm basically taking the pre-load approach in #424 - the tradeoff then is if you want to use a datasource in your context, it'll be pre-loaded.

ArcticSnowman:
True it loads them as start time..   I was using helm template as a general purpose template processor but they made a change recently that assumed the output was YAML and force some YAML lint standard. So I came across you tool and I am converting over to it...

For now I simple load the datasource into a variable at the start of the templates and that allow me to fake the values stored in a context..

```
{{- $values := datasource "values" -}} 
```

assayire:
> True it loads them as start time.. I was using helm template as a general purpose template processor but they made a change recently that assumed the output was YAML and force some YAML lint standard. So I came across you tool and I am converting over to it...
> 
> For now I simple load the datasource into a variable at the start of the templates and that allow me to fake the values stored in a context..
> 
> ```
> {{- $values := datasource "values" -}} 
> ```

This saved my day. Thank you so much.

[...]

PR #423, opened 2018-11-12
add support for default-datasource

moshloop:


hairyhenderson:
@moshloop Thanks for the contribution! Can you please explain your use-case, and why you think this change is necessary?

It seems like this may be also achievable by implementing #404, which I've started working on.

moshloop:
My use case is templating out a bunch of files using variables provided at runtime. e.g. Apache load balancing members etc..

The template shouldn't care where the data is coming from, just that there is data to merge in. 

#404 does get a little of the way there, but would still need to include the arbitrary data source name for every invocation

i.e. we want these to be equivalent:

```
$ gomplate -d config=some:///config/source -i '{{ (ds "config").value }}'
$ gomplate -d config=some:///config/source -i '{{ .value }} --default-datasource config'
```

hairyhenderson:
One way this can be accomplished right now without any code changes is to use nested external templates:

```console
$ echo '{{ .value }}' > in.t
$ echo 'value: hello world' > config.yml
$ gomplate --template t=in.t --datasource config=config.yml --in '{{ template "t" (ds "config") }}'
hello world
```

The input string would remain the same, but the external template (`--template`/`-t`) would change.

hairyhenderson:
@moshloop I think #424 supersedes this.

moshloop:
@hairyhenderson thanks - #424 works as well

PR #424, opened 2018-11-13
New --context flag for pre-loading datasources into context

### LINKED ISSUE #404: Add datasources to root context

hairyhenderson:
Fixes #404 

Adds a new `--context`/`-c` commandline flag to allow pre-loading a datasource into the template's context.

This allows this sort of usage:

```console
$ gomplate -c foo=https://example.com/foo.json -c bar.yml -i '{{ .foo.value }}, {{ .bar.value }}'
foo's value, bar's value
$ gomplate -c .=baz.yml -i '{{ .value }}'
baz's value
```

Caveats:

- all context datasources are loaded first - this can introduce latency
- setting the root (with alias of `.`) makes `.Env` unavailable
- sub-paths can't be given to context datasources in the template (i.e. `{{ datasource "foo" "subpath" }}`)
- inline datasources can't be added to the context at runtime

Other:

- context datasources are still available by name as regular datasources:
    ```console
    $ gomplate -c .=baz.yml -i '{{ .value }} - {{ (ds ".").value }}'
    baz's value - baz's value
    ```

### DIFF
--- a/cmd/gomplate/main.go
+++ b/cmd/gomplate/main.go
@@ -113,6 +113,8 @@ func initFlags(command *cobra.Command) {
 	command.Flags().StringArrayVarP(&opts.DataSources, "datasource", "d", nil, "`datasource` in alias=URL form. Specify multiple times to add multiple sources.")
 	command.Flags().StringArrayVarP(&opts.DataSourceHeaders, "datasource-header", "H", nil, "HTTP `header` field in 'alias=Name: value' form to be provided on HTTP-based data sources. Multiples can be set.")
 
+	command.Flags().StringArrayVarP(&opts.Contexts, "context", "c", nil, "pre-load a `datasource` into the context, in alias=URL form. Use the special alias `.` to set the root context.")
+
 	command.Flags().StringArrayVarP(&opts.InputFiles, "file", "f", []string{"-"}, "Template `file` to process. Omit to use standard input, or use --in or --input-dir")
 	command.Flags().StringVarP(&opts.Input, "in", "i", "", "Template `string` to process (alternative to --file and --input-dir)")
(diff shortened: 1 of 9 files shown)


vitalka200:
Hi @hairyhenderson,
when this PR can be merged.
I wanted to use gomplate to pre-render Helm Charts, and loading datasources is one that is missing.

Thank you!

hairyhenderson:
@vitalka200 Soon! 😉

Keep in mind that gomplate is _not_ a drop-in replacement for Helm. Helm and gomplate have different sets of functions that are supported, and I don't intend for gomplate to ever be fully Helm-compatible.

If you want true pre-rendering for Helm, I suggest using `helm`'s `--dry-run` and `--debug` options.

PR #450, opened 2018-12-20
Document suppressing newlines.

carlosonunez:
I found myself writing multi-line action blocks to improve readability but could not understand why each action left a newline behind. Because I was running this in Docker Compose, I didn't want to alter the `ENTRYPOINT` to your image to do some data munging with `bash` to workaround it.

I saw that some example code in your documentation had leading and trailing minus signs after and before the action delimiters. I figured this was worth something, so I used it in my code and discovered that this is how you suppress newlines.

While this is well-known for those that use Go templates, it wasn't clear to this not-yet-a-Golang dev. I added some documentation to explain this better for folks like me.

I love this tool! Thank you for making it!

### DIFF
--- a/docs/content/syntax.md
+++ b/docs/content/syntax.md
@@ -51,7 +51,7 @@ Hello,  baz!
 
 This might not be desirable.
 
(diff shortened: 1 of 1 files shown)


carlosonunez:
Agreed! I thought of doing this but didn’t know where to put it. Hang on.

carlosonunez:
Done!

hairyhenderson:
@carlosonunez thanks again for this! I'm glad gomplate is useful for you 🙂

PR #496, opened 2019-02-19
Add functions for KMS operations

### LINKED ISSUE #342: AWS KMS Support
I think it'd be cool to have KMS support in `gomplate` allowing "on-the-fly" encryption while templating values. 

The idea's inspired by a tool called [Shush](https://github.com/realestate-com-au/shush) that I came across recently when looking for a way to encrypt sensitive information using AWS KMS.

jbialy:
Fixes #342

🚧 Still WIP but comments / suggestions welcome! This implementation somewhat borrows and was inspired by [shush](https://github.com/realestate-com-au/shush).

🆕 This PR introduces two new AWS functions, KMSEncrypt and KMSDecrypt. 

The intention is to provide basic encryption using AWS KMS. Note that this functionality currently does not support [encryption-context](https://docs.aws.amazon.com/kms/latest/developerguide/encryption-context.html), support for this can be extended in the future.

💻 Usage:

```
export CIPHER=$(bin/gomplate -i '{{ aws.KMSEncrypt "alias/gomplate" "hello world" }}')
```

```
bin/gomplate -i '{{ env.Getenv "CIPHER" | aws.KMSDecrypt }}'
```


aws/ec2info.go
```diff
@@ -39,7 +39,7 @@ type InstanceDescriber interface {
 	DescribeInstances(*ec2.DescribeInstancesInput) (*ec2.DescribeInstancesOutput, error)
 }
 
-// GetClientOptions - Centralised reading of AWS_TIMEOUT
+// GetClientOptions - Centralized reading of AWS_TIMEOUT
```
hairyhenderson: Unnecessary - this is valid (British) spelling 😉

funcs/aws.go
```diff
@@ -30,14 +30,18 @@ func AWSFuncs(f map[string]interface{}) {
 	f["ec2dynamic"] = AWSNS().EC2Dynamic
 	f["ec2tag"] = AWSNS().EC2Tag
 	f["ec2region"] = AWSNS().EC2Region
+	f["kmsEncrypt"] = AWSNS().KMSEncrypt
+	f["kmsDecrypt"] = AWSNS().KMSDecrypt
```
hairyhenderson: can we leave the aliases out for now? I don't think `aws.KMSEncrypt`/`aws.KMSDecrypt` are too cumbersome 🙂

funcs/aws.go
```diff
@@ -65,6 +69,18 @@ func (a *Funcs) EC2Tag(tag string, def ...string) (string, error) {
 	return a.info.Tag(tag, def...)
 }
 
+// KMSEncrypt -
+func (a *Funcs) KMSEncrypt(keyID, plaintext string) (string, error) {
```
hairyhenderson: args should be of type `interface{}` and should be converted inside the function (use `conv.ToString`)

aws/kms.go
```diff
@@ -0,0 +1,51 @@
+package aws
+
+import (
+	"encoding/base64"
```
hairyhenderson: Can you use `github.com/hairyhenderson/gomplate/base64` instead? Keeps things consistent (especially with the `Decode` case)

hairyhenderson:
@jbialy This is awesome! Thanks so much for the contribution 🙂 Aside from some tests (LMK if you want a hand with those), this looks more-or-less ready. I have a few other comments that I've made in-line too 😉

hairyhenderson:
We'll need to add docs for this, but I'll take that on as a follow-up once this is merged (this is more a _note to self_)

funcs/aws.go
jbialy: Sound good, thanks for the tip!

aws/kms.go
jbialy: Yup, I was actually debating if I should use your implementation instead. Makes sense! 😄

jbialy:
> @jbialy This is awesome! Thanks so much for the contribution 🙂 Aside from some tests (LMK if you want a hand with those), this looks more-or-less ready. I have a few other comments that I've made in-line too 😉

Great, thanks for the feedback. Tests are definitely coming! Once ready I'll make the PR "ready for review"!

aws/kms.go
```diff
@@ -0,0 +1,54 @@
...
+	input := &kms.EncryptInput{
+		KeyId:     &keyID,
+		Plaintext: []byte(plaintext),
+	}
+	output, err := k.Client.Encrypt(input)
+	if err != nil {
+		return "", err
+	}
+	ciphertext, err := b64.Encode(output.CiphertextBlob)
+	if err != nil {
```
jbialy: @hairyhenderson Because the `base64.Encode` function can potentially return an error, I had to handle it here. I am however wondering if this is necessary at all, is there a case when the gomplate `base64.Encode` would actually return an error?

https://github.com/hairyhenderson/gomplate/blob/2bab08cea50a3130d2e6fea85589bf2d1bf173d1/base64/base64.go#L8-L10

hairyhenderson: 🤔 😂 clearly not... Looks like that may have gotten caught up in the mega-PR that was #380. It may have made sense to me at the time as a hedge against some future change. This _may_ fulfil some interface too, though I can't think which. And maybe it was just to "balance" out the `Decode` function, which can and does throw `error`s.

_However,_ removing the `error` from `Encode` will break API, so I'm inclined to leave it here. And so, not knowing whether there ever _will_ be an `error`, I think your error-handling code should remain 😉.

aws/kms.go
```diff
@@ -0,0 +1,54 @@
...
+		return "", err
+	}
+	return ciphertext, nil
+}
+
+// Decrypt a base64 encoded cyphertext
```
hairyhenderson: nitpick:

```suggestion
// Decrypt a base64 encoded ciphertext
```

jbialy:
After merging the new changes from the master branch, seems that the test coverage got affected a little bit.. `73% (-0.2% change)`.

hairyhenderson:
@jbialy don't worry about the coverage - 0.2% is nothing 😉

PR #525, opened 2019-03-27
Fixing random test that fails rarely 🎲

hairyhenderson:
Just a simple logic bug in the unit test... Ran into it during a recent PR! 😂🎲


hairyhenderson:
🤔 CircleCI's having an outage... I'm going to merge this without it 🤠

Issue #546, opened 2019-05-02
Switch from gometalinter to golang-ci


nexeck:
gometalinter seems to be deprecated https://github.com/alecthomas/gometalinter/issues/590
They suggest to use https://github.com/golangci/golangci-lint

hairyhenderson:
Thanks for the heads up! I'm at a conference this week so won't have much time to look at the PR, but will definitely look at it.

nexeck:
https://github.com/hairyhenderson/dockerfiles/pull/55 This PR is related to the failing Circle CI Lint test.

hairyhenderson:
Fixed by #547 - thanks again @nexeck!

PR #547, opened 2019-05-02
chore: Switch from gometalinter to golangci-lint

nexeck:
Github Issue: #546

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,39 +117,32 @@ gomplate.png: gomplate.svg
 	cloudconvert -f png -c density=288 $^
 
 lint:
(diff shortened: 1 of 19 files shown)


nexeck:
I do not know how circle ci works. But it seems that the provided golangci-lint version is not the newest.

nexeck:
Please see: https://github.com/hairyhenderson/dockerfiles/pull/55 for a fix for the lint error.

hairyhenderson:
Thanks @nexeck! I'm going to add some commits to this PR - `golangci-lint` is significantly faster than `gometalinter` was, and so I'm going to remove the `slow-lint` target.

Issue #584, opened 2019-08-16
Add POSIX-shell-compatible quote function


charles-dyfis-net:
At the cost of adding new dependencies, this could be `shellquote.Join()` from https://github.com/kballard/go-shellquote, or `shellescape.Quote()` from https://github.com/alessio/shellescape; both are MIT-licensed.

hairyhenderson:
Thanks @charles-dyfis-net - this sounds like a good idea! I'm thinking `strings.ShellQuote` may be the right function name.

I like the implementation in github.com/alessio/shellescape far better - the other one seems a bit convoluted! I think you offered to issue a PR for this, it would be most welcome! If you want to simply copy/paste to avoid the extra dependency, that'd be fine too. Either way is good!

hairyhenderson:
Closed in #587

PR #587, opened 2019-08-19
Implement shell-compatible quoting

charles-dyfis-net:
As discussed in #584. Ended up implementing inline as I don't trust the regex-based approach to deciding which strings need to be quoted to be safe for all possible extended shells.

### DIFF
--- a/docs-src/content/functions/strings.yml
+++ b/docs-src/content/functions/strings.yml
@@ -252,6 +252,24 @@ funcs:
       - |
         $ echo 'Rock & Roll @ Cafe Wha?' | gomplate -d in=stdin: -i '{{ strings.Slug (include "in") }}'
         rock-and-roll-at-cafe-wha
+  - name: strings.ShellQuote
+    alias: shellQuote
+    description: |
+      Given a string, emits a version of that string that will evaluate to its literal data when expanded by any POSIX-compliant shell.
+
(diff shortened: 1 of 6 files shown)


hairyhenderson:
This looks great @charles-dyfis-net - thank you! Can you move the implementation & tests into the `github.com/hairyhenderson/gomplate/strings` package? Then just reference it from `funcs/strings.go` as `gompstrings.ShellQuote`

charles-dyfis-net:
I'm a little unclear on how to do that cleanly.

As I read the code, the convention is for `gompstrings` to have logic that isn't aware of the type-conversion work necessary to be used in template functions live there. However, the bulk of the `func.strings` logic, or at least, the logic that requires aggressive testing, *is* type conversion.

Is e8a9393dd6ebba69b3d00ef81a745a7b421f87c4 what you were looking for?

hairyhenderson:
> As I read the code, the convention is for `gompstrings` to have logic that isn't aware of the type-conversion work necessary to be used in template functions live there. However, the bulk of the `func.strings` logic, or at least, the logic that requires aggressive testing, _is_ type conversion.

You're exactly right - and what you did in 8173e67 is perfect 😉 - the idea is mostly that I want the function to be usable programatically (and the `gomplate/functions` package is intended to be treated as internal).

hairyhenderson:
Thanks again for this, @charles-dyfis-net!

PR #598, opened 2019-10-04
Moving to go modules

### LINKED ISSUE #545: Switch from dep to go mod
Is there a plan to switch from dep to go mod?

hairyhenderson:
Fixes #545 

This also involves a bunch of version bumps in dependencies, largely because some of them were locked at versions that just plain didn't work with modules (weird ambiguities and such).


hairyhenderson:
Holy cow, it worked! And it only took changes to 774 files 😂

go.mod
```diff
@@ -0,0 +1,45 @@
...
+	github.com/zealic/xignore v0.3.3
+	golang.org/x/crypto v0.0.0-20191002192127-34f69633bfdc
+	golang.org/x/net v0.0.0-20191003171128-d98b1b443823 // indirect
+	golang.org/x/sys v0.0.0-20191003212358-c178f38b412c // indirect
+	golang.org/x/text v0.3.2 // indirect
+	golang.org/x/time v0.0.0-20190921001708-c4c64cad1fd0 // indirect
+	gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405
+	gopkg.in/yaml.v3 v3.0.0-20190924164351-c8b7dadae555
+	gotest.tools/v3 v3.0.0
+	k8s.io/client-go v11.0.0+incompatible
```
fatih: Why do you pull in k8s.io's client?

hairyhenderson: that's a good question 😲... I'll dig into that tomorrow morning 🛌

hairyhenderson: oh actually, that's a simple answer - it's here: https://github.com/hairyhenderson/gomplate/blob/7724a8d8e6b0437b151cbfc5f4ca2ee5695d0bf8/coll/jsonpath.go#L7

fatih: I see. I think this is a very big dependency for such a small package. I wonder if you could look into this to make this package more leaner. Thanks already for moving to Go modules and I appreciate. I think the less dependencies the better. You or any other user, they shouldn't pull down a big dependency like this.

hairyhenderson: @fatih maybe I'm misunderstanding how go modules work, but is all of `k8s.io/client-go` downloaded just to get that one package? When I `go mod vendor` it, only that package is included:

```
$ du -h vendor/k8s.io/client-go
40K     vendor/k8s.io/client-go/util/jsonpath
40K     vendor/k8s.io/client-go/util
20K     vendor/k8s.io/client-go/third_party/forked/golang/template
20K     vendor/k8s.io/client-go/third_party/forked/golang
20K     vendor/k8s.io/client-go/third_party/forked
20K     vendor/k8s.io/client-go/third_party
72K     vendor/k8s.io/client-go
```

I am seeing that when I `go get` it, it pulls down all of `k8s.io/client-go` into my module cache, which is ~6MB and 1000+ files 🤔...

PR #625, opened 2019-10-13
Supporting s3 datasources

### LINKED ISSUE #566: S3 Datasource or Functions
It would be great if `gomplate` could use S3 as a datasource, or had functions to read objects from S3, so that they could be used in templates!

hairyhenderson:
New support for S3 datasources! This uses the Go CDK (https://gocloud.dev) to make life (in theory) a bit easier. This should also pave the way for other blob storage types with relative ease.

Fixes #566 


hairyhenderson:
I've tested this out a few ways with both public and private buckets, and it seems to be working as expected. Time to merge!

PR #641, opened 2019-10-20
Placed a newline before the start of a list

Ilyes512:
Placed a (missing) newline before the start of a list so it renders the list good.

Currently shown as:
Screenshot 2019-10-20 at 16 24 44


Ilyes512:
There is one more render problem on the same page:

Screenshot 2019-10-20 at 16 38 10

hairyhenderson:
Thanks @Ilyes512! I appreciate that - I thought I'd caught all those a while back, but clearly not 😂

hairyhenderson:
> There is one more render problem on the same page:

Thanks - fixed that one too.

PR #808, opened 2020-04-11
Produce multi-platform Docker images

hairyhenderson:
An attempt to produce images for Linux on amd64, arm64, arm v6/v7, and Windows on amd64.

Uses experimental Dockerfile syntax with BuildKit/`docker buildx`. This means I probably won't be able to build with DockerHub auto-builds longer-term.


hairyhenderson:
(closing in favour of #811)

PR #811, opened 2020-04-19
Produce multi-platform Docker images

hairyhenderson:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,9 +4,104 @@ on: [push]
 jobs:
   docker-build:
     runs-on: ubuntu-latest
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - '5000:5000'
+    env:
+      DOCKER_BUILDKIT: 1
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -4,9 +4,104 @@ on: [push]
...
+        docker buildx ls
+        docker ps
+        docker images
+    - run: docker info && docker version
     - uses: actions/checkout@v1
-    - name: Make Docker images
-      run: make docker-images
+    - name: determine if this is a tag
+      run: |
+        if (git describe --abbrev=0 --exact-match &>/dev/null); then
```
glensc, line 39: you don't need subshell here:

```
        if git describe --abbrev=0 --exact-match &>/dev/null; then
```

hairyhenderson: thanks @glensc - this is a holdover from many copy-pastes ago... I'll happily merge a PR if you want to make the change 😉

glensc: no worries. you can leave as is as well. totally up to you.

PR #819, opened 2020-04-23
Fix broken file datasource URL parsing (query string was ignored)

### LINKED ISSUE #809: type query not parsed by defineDataSource
Brief: Load a [.env] file and export it as yaml:

[.env]: https://docs.gomplate.ca/datasources/#the-env-file-format

```yaml
{{- defineDatasource "env_file" ".env" -}}

config:
  {{ (ds "env_file") | data.ToYAML | strings.Indent 1 "  " | trimSpace }}
```

however, my local file is named `.env.foo`, there seems no way to specify [mime-type]:

[mime-type]: https://docs.gomplate.ca/datasources/#overriding-mime-types

```
{{- defineDatasource "env_file" ".env.staging?type=application/x-env" -}}
```

I tried also the absolute path and `file://` prefix as well.

```
config:
  template: values.yml:4:6: executing "values.yml" at <ds "env_file">: error calling ds: Couldn't read datasource 'env_file': Can't stat /tmp/env-import/.env.staging?type=application/x-env: stat /tmp/env-import/.env.staging?type=application/x-env: no such file or directory
```

hairyhenderson:
Fixes #809

Turns out paths passed on the commandline would've also suffered from this bug - the query string on non-absolute URLs (i.e. file paths) wasn't getting parsed.

So this wouldn't have been recognized:

```console
$ gomplate -d foo=./foo?type=application/json ...
```

And neither would inline uses of `defineDatasource` (such as reported in #809):

```
{{- defineDatasource "env_file" ".env.staging?type=application/x-env" -}}
```

Incidentally, this also fixes an issue I've experienced before when trying to emulate reads from Vault with a local directory:

```console
$ gomplate -d secrets=./secret_dir/?type=application/json -i '{{ (ds "secrets" "some/key").value }}'
```

### DIFF
--- a/data/datasource.go
+++ b/data/datasource.go
@@ -230,7 +230,7 @@ func parseSource(value string) (source *Source, err error) {
 			err = errors.Errorf("Invalid datasource (%s). Must provide an alias with files not in working directory", value)
 			return nil, err
 		}
-		source.URL, err = absURL(f)
+		source.URL, err = absFileURL(f)
 		if err != nil {
 			return nil, err
 		}
@@ -272,7 +272,7 @@ func parseSourceURL(value string) (*url.URL, error) {
 	}
 
 	if !srcURL.IsAbs() {
-		srcURL, err = absURL(value)
+		srcURL, err = absFileURL(value)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


glensc:
thanks!

PR #891, opened 2020-07-10
Add gcp.Meta function, equivalent to aws.EC2Meta but for GCP

jen20:
This commit adds a new namespace and function: `gcp.Meta`, which can be used to look up values from the GCP Instance Metadata service.

An example usage:

```
echo '{{ gcp.Meta "id" }}' | gomplate
```

This also supports paths, so usage like this works:

```
echo '{{ gcp.Meta "network-interfaces/0/ip" }}' | gomplate
```

I've added a new documentation page to cover the GCP namespace, but it's unclear whether I have missed a spot where this page needs adding to a layout or similar.

### DIFF
--- a/docs-src/content/functions/gcp.yml
+++ b/docs-src/content/functions/gcp.yml
@@ -1,0 +1,36 @@
+ns: gcp
+preamble: |
+  The functions in the `gcp` namespace interface with various Google Cloud Platform
+  APIs to make it possible for a template to render differently based on the GCP
+  environment and metadata.
+
+  ### Configuring GCP
+
+  A number of environment variables can be used to control how gomplate communicates
+  with GCP APIs.
+
+  | Environment Variable | Description |
+  | -------------------- | ----------- |
+  | `GCP_META_ENDPOINT` | _(Default `http://metadata.google.internal`)_ Sets the base address of the instance metadata service. |
+  | `GCP_TIMEOUT` | _(Default `500`)_ Adjusts timeout for API requests, in milliseconds. |
+funcs:
+  - name: gcp.Meta
+    description: |
+      Queries GCP [Instance Metadata](https://cloud.google.com/compute/docs/storing-retrieving-metadata) for information.
+
+      For times when running outside GCP, or when the metadata API can't be reached, a `default` value can be provided.
+    pipeline: false
+    arguments:
+      - name: key
+        required: true
+        description: the metadata key to query
+      - name: default
+        required: false
+        description: the default value
+    examples:
+      - |
+        $ echo '{{gcp.Meta "id"}}' | gomplate
+        1334999446930701104
+      - |
+        $ echo '{{gcp.Meta "network-interfaces/0/ip"}}' | gomplate
+        10.128.0.23
--- a/docs/content/functions/gcp.md
+++ b/docs/content/functions/gcp.md
@@ -1,0 +1,50 @@
+---
+title: gcp functions
+menu:
+  main:
+    parent: functions
+---
+
+The functions in the `gcp` namespace interface with various Google Cloud Platform
+APIs to make it possible for a template to render differently based on the GCP
+environment and metadata.
+
+### Configuring GCP
+
+A number of environment variables can be used to control how gomplate communicates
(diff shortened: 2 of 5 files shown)


hairyhenderson:
Thanks @jen20, this is awesome! 😬

Looks like `make lint` is failing - can you clear those up?

> I've added a new documentation page to cover the GCP namespace, but it's unclear whether I have missed a spot where this page needs adding to a layout or similar.

Function docs are generated from the YAML files in `/docs-src/content/functions/` - probably the best approach is to copy from one of the more recently-modified ones and alter as required. If you run `make docs/content/functions/gcp.md` it should generate in the right format. I can also handle this for you if you'd like.

hairyhenderson:
Also I've been thinking that some of the `aws.*` functions might have been better implemented as datasources.

What do you think of something like a `gcp+meta://` datasource? I'm not really sure either way, but I'd value your input!

jen20:
Hi! I've fixed up the lint errors from `golangci-lint`. I've also looked into the CodeClimate ones and am quite unconvinced that it would be beneficial to fix any of the issues flagged there (particularly the largely duplicated `GetClientOptions` function).

I'll take another look over the docs now.

> What do you think of something like a gcp+meta:// datasource?

I'm not especially opposed to a data source, but personally I find them more difficult to work with than plain functions, and having a different approach between major cloud vendors seems suboptimal at the moment. Perhaps this is something that could be looked at for a major version bump in future along with AWS?

hairyhenderson:
> I'm not especially opposed to a data source, but personally I find them more difficult to work with than plain functions, and having a different approach between major cloud vendors seems suboptimal at the moment. Perhaps this is something that could be looked at for a major version bump in future along with AWS?

Yeah, I'd probably want to add an `aws+meta://` datasource as well. And it definitely wouldn't mean taking away the existing `aws` functions. My thought was that it would be easier to reuse templates between different providers that way. I'll give it some more thought - there's some functions in the `sockaddr` namespace that may benefit from being datasource-ized as well. Or maybe not! It may well be more trouble than it's worth 😉

> I've also looked into the CodeClimate ones and am quite unconvinced [...]

Yeah that's often my attitude towards CodeClimate 😂 (though it _does_ catch dumb stuff on occasion). I'll make them disappear.

jen20:
> My thought was that it would be easier to reuse templates between different providers that way.

Yes, I suppose you are right here - the tradeoff is that the invocation becomes more complex. Having both options available seems like it might give the best of both worlds there.

>  I'll make them disappear.

Great, thanks. Is there anything else that needs addressing by me to land this? If not, is there any particular date you have in mind for a release - no worries if not, but it would avoid me some repackaging work to run a custom build in the meantime ;-)

docs-src/content/functions/gcp.yml
```diff
@@ -0,0 +1,35 @@
...
+  environment and metadata.
+
+  ### Configuring GCP
+
+  A number of environment variables can be used to control how gomplate communicates
+  with GCP APIs.
+
+  | Environment Variable | Description |
+  | -------------------- | ----------- |
+  | `GCP_TIMEOUT` | _(Default `500`)_ Adjusts timeout for API requests, in milliseconds. |
```
hairyhenderson, line 14: Can you document `GCP_META_ENDPOINT` here too?

hairyhenderson:
> Is there anything else that needs addressing by me to land this?

Just the one doc nitpick (`GCP_META_ENDPOINT`)

> is there any particular date you have in mind for a release

No particular date, but I have some other functions I want to get in before I tag v3.8.0. The `hairyhenderson/gomplate:latest` image will reflect the git `master` branch once it's merged, so you should be able to avoid repackaging in the meantime! I'm hoping to finish the other stuff up within a week or so (modulo $DAYJOB), so it shouldn't be _too_ long!

jen20:
> Just the one doc nitpick (GCP_META_ENDPOINT)

Done - I realised the reason I missed this is because `AWS_META_ENDPOINT` is also not documented. I've opened #897 to address this.

hairyhenderson:
Awesome, thanks for this @jen20 😬

Issue #861, opened 2020-06-02
Intermittent failure using -slim binaries, exit code 127


bhartshorn:
Hi, I've recently encountered a problem using the -slim versions of the gomplate release binaries. When running hundreds of times in succession, eventually gomplate will fail and return an exit code `127`. I suspect this is due to an issue with the upx compression used for the -slim binary.

## Tested versions:
I'm going to try to test some other combinations and will report back

All of these combinations failed when run in rapid succession using the script shared below:
- (v3.6.0) gomplate_linux-amd64-slim on Ubuntu Xenial 16.04 (Docker) and RHEL 7.7 (Metal)
- (v3.7.0) gomplate_linux-amd64-slim on RHEL 7.7 (Metal)

These combinations completed 100,000 iterations without failure:
- (v3.6.0) gomplate_linux-amd64 on RHEL 7.7 (Metal)

## Steps to reproduce

I've managed to reproduce the issue in a docker container, hopefully you can reproduce it this way also.
1. Files:
```
.
├── Dockerfile
├── gomplate_linux-amd64      (v3.6.0)
├── gomplate_linux-amd64-slim (v3.6.0)
└── test.sh
```

2. Dockerfile:
```
FROM ubuntu:xenial
COPY test.sh .
COPY gomplate* /usr/bin/
```

3. test.sh:
```
#!/usr/bin/env bash
set -e
#set -x

if [[ $# -eq 1 ]]; then
  bin="gomplate_linux-amd64-$1"
else
  bin="gomplate_linux-amd64"
fi

echo "Running ${bin}"

for ((ITER = 1; ITER <= 100000; ITER++)); do
  export ITER=${ITER}
  [ $(($ITER % 100)) != "0" ] && echo -n "." || echo " Running ${ITER}"
  avar=avar bvar=bvar ${bin} '--in={ "avar": {{ .Env.avar | toJSON }}, "bvar": {{ .Env.bvar | toJSON }} }' --out=output.yaml >gomplate.log 2>&1 || {
    echo -e "\nexit code: $?"
    echo "Ended on iteration ${ITER}"
    break
  }
  rm -f gomplate.log output.yaml
done
```

4. Docker build:
`> docker build . -t gomplate-upx-issue:latest`

5. Run in Docker:
```
> docker run -it --entrypoint /bin/bash gomplate-upx-issue
root@496699bf7dda:/# ./test.sh slim
Running  /usr/bin/gomplate_linux-amd64-slim
[...]

PR #1003, opened 2020-12-15
Updates upx to 3.96

alepar:
**Motivation**
I'm having issues running gomplate's docker image on arm64. The binary failes to execute properly:
```
/ # /bin/gomplate
PROT_EXEC|PROT_WRITE failed.
```

I've traced this down to upx. If I decompress the binary, it works just fine.
```
/ # /tmp/upx/upx-3.96-arm64_linux/upx -d /bin/gomplate 
                       Ultimate Packer for eXecutables
                          Copyright (C) 1996 - 2020
UPX 3.96        Markus Oberhumer, Laszlo Molnar & John Reiser   Jan 23rd 2020

        File size         Ratio      Format      Name
   --------------------   ------   -----------   -----------
  35061760 <-   6503964   18.55%   linux/arm64   gomplate

Unpacked 1 file.
/ # /bin/gomplate
^C
/ # /bin/gomplate --version
gomplate version 3.8.0
```

Moreover, if I recompress the binary with the original cli options (I believe you're using '--lzma'), it still works. I'm using upx 3.96 though, so I believe there was probably a bug that they fixed between 3.94 and 3.96, that is causing this.
Can't really confirm if compressing with upx 3.94 will reproduce the issue, since 3.94 doesn't seem to have an arm64 binary yet :(  But I'm still hoping this PR will fix it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.1.7-experimental
-FROM --platform=linux/amd64 hairyhenderson/upx:3.94 AS upx
+FROM --platform=linux/amd64 hairyhenderson/upx:3.96 AS upx
 
 FROM --platform=linux/amd64 golang:1.15.0-alpine3.12 AS build
 


hairyhenderson:
Hi @alepar - I've had a number of issues with UPX between versions 3.94, 3.95, and 3.96. Some of them are documented in https://github.com/upx/upx/issues/222.

I actually moved up to UPX 3.96 for gomplate 3.7.0, but then moved back down to 3.8.0, due to a regression - #861. Downgrading happened in #875.

Unfortunately at this point the UPX project seems to be progressing very slowly, and the developers have not been responding well to bug reports.

I think the only thing that I can recommend to you at this point is to use the regular (non `-slim`) builds of gomplate. It also turns out that while the non-compressed binaries are smaller, they execute _much_ faster.

alepar:
I see, I wasn't aware of the history there, thanks for the explanation!

Yeah, I've switched to `hairyhenderson/gomplate:stable` now. After, I have to admit, a bit of a struggle, which was mainly due to my own stupidity, I got it working.

I didn't seem to find any published docker images that would have uncompressed binary and contained a shell at the same time, which was an inconvenience, but only a minor one. Maybe I just didn't look close enough...

Anyways, thank you for your work here!

hairyhenderson:
> I didn't seem to find any published docker images that would have uncompressed binary and contained a shell at the same time, which was an inconvenience, but only a minor one. Maybe I just didn't look close enough...

That's a good point - the `alpine` variant uses the compressed binary. I will change that in the next release as this isn't the first time it's tripped someone up 😉

PR #1025, opened 2021-01-02
Major refactoring around datasources

hairyhenderson:
This is a WIP... But mostly it's around abstracting the API around data requests so the code can be shared.

There's still a bunch of work to do, and I may have changed a bit _too much..._ The idea is not to break API, though that may be inevitable.


hairyhenderson:
Stale, but still worth keeping around - likely to be superseded by #1336

PR #1084, opened 2021-03-02
Support Vault Kubernetes Auth

spearki:
This PR adds support for authenticating to Vault with a Kubernetes service account token. As it happens implementing this also makes it trivial to support vanilla JWT auth, so that is included here as well.

There's an integration test for both Kubernetes and JWT - I have faked the Kubernetes token review API in the tests. This required tightening up the certificate creation code a bit since Vault is quite strict about the contents of the cert.


hairyhenderson:
Hi @conkiztador, thank you for the contribution, and sorry it's taken me so long to get to it!

I'll take a bit of a deeper look in a bit, but first can you rebase the PR?

There's also a new `internal/tests/inprocess-integration` directory which contains duplicates of the tests in `internal/tests/integration`, except that they're run in-process instead of spawning the binary. Can you duplicate your tests there too please? Eventually I hope to winnow it down to a single directory, but that'll take time which is in short supply at the moment 😉

vault/auth.go
```diff
@@ -232,3 +236,49 @@ func (v *Vault) TokenLogin() (string, error) {
...
+	}
+	mount := env.Getenv("VAULT_AUTH_JWT_MOUNT", "jwt")
+	return v.jwtLogin(mount, token, role)
+}
+
+func (v *Vault) KubernetesLogin() (string, error) {
+	if os.Getenv("VAULT_AUTH_KUBERNETES_TOKEN_FILE") == "" {
+		os.Setenv("VAULT_AUTH_KUBERNETES_TOKEN_FILE", kubernetesTokenPath)
+		defer os.Unsetenv("VAULT_AUTH_KUBERNETES_TOKEN_FILE")
+	}
```
hairyhenderson, line 273: I was a little confused why you used `os.Getenv` here instead of `env.Getenv`, until I realized what's going on. This would benefit from an explanatory comment, because I _know_ I'll get confused again 😉

```suggestion
	// allow env.Getenv to fall back to the default token path file
	if os.Getenv("VAULT_AUTH_KUBERNETES_TOKEN_FILE") == "" {
		os.Setenv("VAULT_AUTH_KUBERNETES_TOKEN_FILE", kubernetesTokenPath)
		defer os.Unsetenv("VAULT_AUTH_KUBERNETES_TOKEN_FILE")
	}
```

(maybe you can think of a better way to word it, but I think that's good enough)

docs/content/datasources.md
```diff
@@ -698,6 +698,8 @@ This table describes the currently-supported authentication mechanisms and how t
...
 | [`aws`](https://www.vaultproject.io/docs/auth/aws.html) | The env var  `$VAULT_AUTH_AWS_ROLE` defines the [role](https://www.vaultproject.io/api/auth/aws/index.html#role-4) to log in with - defaults to the AMI ID of the EC2 instance. Usually a [Client Nonce](https://www.vaultproject.io/docs/auth/aws.html#client-nonce) should be used as well. Set `$VAULT_AUTH_AWS_NONCE` to the nonce value. The nonce can be generated and stored by setting `$VAULT_AUTH_AWS_NONCE_OUTPUT` to a path on the local filesystem.<br/>If the back-end is mounted to a different location, set `$VAULT_AUTH_AWS_MOUNT`.|
+| [`jwt`](https://www.vaultproject.io/docs/auth/jwt) | `$VAULT_AUTH_JWT_TOKEN` and `$VAULT_AUTH_JWT_ROLE` must be set.<br/> If the back-end is mounted to a different location, set `$VAULT_AUTH_JWT_MOUNT`. |
+| [`kubernetes`](https://www.vaultproject.io/docs/auth/kubernetes) | `$VAULT_AUTH_KUBERNETES_ROLE` must be set to the role to log into vault with. Gomplate will authenticate using the Pod service account token by default. To use a different token, set `$VAULT_AUTH_KUBERNETES_TOKEN`.<br/> If the back-end is mounted to a different location, set `$VAULT_AUTH_KUBERNETES_MOUNT`. |
```
hairyhenderson, line 702: These are great, but should be in order of precedence - currently your implementation has the EC2 (`aws`) role last. I think that's the right order, so these docs just need to change. Also I've fixed just a couple minor typos:

```suggestion
| [`jwt`](https://www.vaultproject.io/docs/auth/jwt) | `$VAULT_AUTH_JWT_TOKEN` and `$VAULT_AUTH_JWT_ROLE` must be set.<br/>If the back-end is mounted to a different location, set `$VAULT_AUTH_JWT_MOUNT`. |
| [`kubernetes`](https://www.vaultproject.io/docs/auth/kubernetes) | `$VAULT_AUTH_KUBERNETES_ROLE` must be set to the role to log into Vault with. Gomplate will authenticate using the Pod's service account token by default. To use a different token, set `$VAULT_AUTH_KUBERNETES_TOKEN`.<br/>If the back-end is mounted to a different location, set `$VAULT_AUTH_KUBERNETES_MOUNT`. |
| [`aws`](https://www.vaultproject.io/docs/auth/aws.html) | The env var  `$VAULT_AUTH_AWS_ROLE` defines the [role](https://www.vaultproject.io/api/auth/aws/index.html#role-4) to log in with - defaults to the AMI ID of the EC2 instance. Usually a [Client Nonce](https://www.vaultproject.io/docs/auth/aws.html#client-nonce) should be used as well. Set `$VAULT_AUTH_AWS_NONCE` to the nonce value. The nonce can be generated and stored by setting `$VAULT_AUTH_AWS_NONCE_OUTPUT` to a path on the local filesystem.<br/>If the back-end is mounted to a different location, set `$VAULT_AUTH_AWS_MOUNT`.|
```

hairyhenderson:
FYI this'll be implemented by #1336 as `go-fsimpl` supports all native Vault auth schemes: https://pkg.go.dev/github.com/hairyhenderson/go-fsimpl@v0.0.0-20230430001005-e7eabc350f7c/vaultfs/vaultauth

PR #1125, opened 2021-04-10
Support URLs for external templates

### LINKED ISSUE #963: Template to process from remote datasource (such as aws+ssm)
I read the documentation but couldn't quite figure the answer to my question.

What I would like to do is to store the complete template in AWS SSM Parameter Store. Inside the template is the datasources etc. This way I can use the standard gomplate docker and keep the template easily updatable.

The template is then output to an application specific config file.

Is there a way to dynamically grab the template itself? Then process the downloaded template.

hairyhenderson:
Fixes #963 

TODO:
- [x] support `file://` URL scheme
- [ ] support non-`file` URL scheme (same set as datasources)
- [ ] update docs


hairyhenderson:
#1410 implemented this, though not all URL schemes are supported quite yet

PR #1283, opened 2022-01-10
Update integration test to work properly with Consul v.11+

hairyhenderson:
Not sure why I didn't see these test errors in CI, but I'm seeing them locally.

```
* Unexpected response code: 410 (Endpoint /v1/acl/create for the legacy ACL system was removed in Consul 1.11.)
```

### DIFF
--- a/internal/tests/integration/datasources_consul_test.go
+++ b/internal/tests/integration/datasources_consul_test.go
@@ -4,11 +4,11 @@
 package integration
 
 import (
-	"encoding/base64"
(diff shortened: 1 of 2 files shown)


hairyhenderson:
> Not sure why I didn't see these test errors in CI

that's because I was using a _very_ old consul version (because the `hairyhenderson/gomplate-ci-build` image build was failing until [just recently](https://github.com/hairyhenderson/dockerfiles/pull/454))

PR #1287, opened 2022-01-20
Add function to list datasources

jeevansai502:


hairyhenderson:
Looks reasonable! Can you please add some documentation for this?

Also, because `d.Sources` is a `map`, the list will be randomly-ordered, which may not be ideal. It might be best to sort the slice alphabetically before returning, for a deterministic result.

hairyhenderson (approved):
LGTM - Thanks @jeevansai502!

PR #1317, opened 2022-02-22
Support piping input to plugin

### LINKED ISSUE #1311: Call an external binary when rendering a template to multiple outputs
I have a use case where I need to render multiple files from one template, which works great as described in #485. However, I found myself in the awkward situation where the rendered files need to be passed through another templating engine, namely [pug](https://pugjs.org/), which isn't possible currently. What I'm thinking of is to have a function called `os.Exec` like this:

```
{{- range $region := slice "AP" "EU" "US" -}}
{{- $ctx := dict "region" $region }}
{{- $outPath := printf "%s.html" $region }}
{{- tmpl.Exec "mytemplate" $ctx | os.Exec "pug" | file.Write $outPath }}
{{- end -}}
```

I looked at `--exec-pipe`, but it doesn't seem intended for such a use case.

While I can use gomplate to render the `.pug` files and then use bash to walk the directory which contains them recursively, it feels less elegant and I'll need to create a temp folder somewhere...

I can try and put together a PR if you're interested in this addition.

PS: Thank you for this wonderful utility!

hairyhenderson:
Fixes #1311 

Instead of only providing the input as the final arg to the plugin command, this allows providing the input as its stdin instead.

### DIFF
--- a/docs/content/config.md
+++ b/docs/content/config.md
@@ -300,22 +300,72 @@ outputMap: |
 
 See [`--plugin`](../usage/#--plugin).
 
-An mapping of key/value pairs to plug in custom functions for use in the templates.
+A map that configures custom functions for use in the templates. The key is the
+name of the function, and the value configures the plugin. The value is a map
+containing the command (`cmd`) and the options `pipe` (boolean) and `timeout`
+(duration).
+
+Alternatively, the value can be a string, which sets `cmd`.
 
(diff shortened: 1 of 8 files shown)


mihaitodor:
That looks awesome @hairyhenderson and it does just what I needed. Thank you so much for implementing it! ❤️

PR #1336, opened 2022-03-10
Use go-fsimpl to read from datasources

### LINKED ISSUE #862: Support Vault's secrets engine v2
Currently the vault datasource will not work with v2 version of Vault's secret engine. Extending the datasource to support this would allow for the use of versioned secrets!

https://www.vaultproject.io/api/secret/kv/kv-v2

### LINKED ISSUE #963: Template to process from remote datasource (such as aws+ssm)

hairyhenderson:
Replace datasource code with [go-fsimpl](https://github.com/hairyhenderson/go-fsimpl).

Note - this will go in v4 since it breaks compatibility by removing support for certain datasource types. It also unifies the URL formats in ways that may be slightly incompatible with v3. My hope is to reduce that impact, but it may not be possible for every datasource type.

- Fixes #963
- Fixes #862


hairyhenderson:
_almost_ there - looks like some Windows compatibility issues still 😬

hairyhenderson:
_Wow_ do I ever dislike dealing with Windows paths...

This is not perfect, but all the tests pass now, so I think it's ready to merge. There will be a bunch of cleanup/refactoring PRs to follow before 4.0 gets cut, but I want to get a prerelease out ASAP to see if I can get these changes tested in the wild...

PR #1350, opened 2022-03-21
Fix merging context/datasources config

paddycarey:
This PR fixes config merging when context/datasources are provided on the command line but _not_ in the config file.

When a config file is used which doesn't contain `context` or `datasources` keys, those values are parsed as `nil`. Then when `config.Config.MergeFrom` is called (and subsequently `config.mergeDataSources`) the application panics as it's trying to write to a nil map.


hairyhenderson:
Thanks @paddycarey!

hairyhenderson:
Sorry for the delay - been having some CI trouble 😅

PR #1402, opened 2022-05-23
Add support for aws imdsv2

### LINKED ISSUE #848: Switch AWS data source to IMDSv2
AWS' instance metadata service has received a more secure way of accessing it (version 2 called IMDSv2): https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/

Looking at https://github.com/hairyhenderson/gomplate/blob/master/aws/ec2meta.go we are probably still calling it with the previous version.

As both versions are supported by AWS the migration plan could be as easy as switching to IMDSv2 instead, which means fetching a shortlived token first.

surki:
Changes:

* Start using the AWS SDK for retrieving instance data (to future proof against changes that might come in EC2 auth)
* Supports both IMDSv1 and IMDSv2 transparently

Fixes #848

### DIFF
--- a/aws/ec2info_test.go
+++ b/aws/ec2info_test.go
@@ -12,8 +12,8 @@ import (
 )
 
 func TestTag_MissingKey(t *testing.T) {
-	server, ec2meta := MockServer(200, `"i-1234"`)
-	defer server.Close()
+	ec2meta := MockEC2Meta(map[string]string{"instance-id": "i-1234"}, "")
+
 	client := DummyInstanceDescriber{
 		tags: []*ec2.Tag{
 			{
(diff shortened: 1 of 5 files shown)


surki:
@hairyhenderson looks like there is some failure in vault ec2 test, will check and fix it.

hairyhenderson:
Hi @surki, thanks for taking the time to work on this!

I'm wondering if you saw the discussion in #848? This could fix things for the v3.11 release (if it can be merged soon!) but in v4 I will be switching to https://github.com/hairyhenderson/go-fsimpl as a back-end so the IMDSv2 support will need to be in there instead.

As for the test failure - I've re-triggered the build as I suspect that test _may_ just be flaky. But it's also potentially a real issue as related code is being modified...

surki:
@hairyhenderson <https://github.com/hairyhenderson> Can you please take a
look at  it? I have fixed the test case failures.

There are few lint failures, one is unrelated to this change, other seems
to be from multiple return statements. Do you want me to fix them as well?

hairyhenderson:
@surki yeah - don't worry about `go-fsimpl`!

If you rebase the CI lint failure should be resolved, then I can merge this.

surki:
> @surki yeah - don't worry about `go-fsimpl`!
> 
> If you rebase the CI lint failure should be resolved, then I can merge this.

Thanks, done

btw, "codeclimate" check is failing, should I fix it?

hairyhenderson:
I rebased again. Will merge when the build is complete. Thanks @surki!

PR #1480, opened 2022-08-13
Fix nested template panic

mihaitodor:
This is a naive attempt to fix a panic that came up on the #gomplate channel on the Gophers Slack. I'm really not sure if this is the best way to fix the issue, but the changes in `TestParseConfigFile` will help you reproduce the issue easily. Alternatively, maybe this custom `UnmarshalYAML` should take the receiver by value rather than by pointer, but I'm not sure if the rest of the custom logic will still work in that case.

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -40,7 +40,6 @@ output: out/`
 	c = &Config{
 		InputDir:  "in/",
 		OutputMap: "{{ .in }}",
-		Templates: []string{},
 	}
 	expected = `input: in/
 output: {{ .in }}`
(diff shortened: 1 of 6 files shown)


hairyhenderson:
Thanks so much for this @mihaitodor!! I'm surprised there wasn't an integration test that caught this - I'll add one of those too.

I'll be glad to get rid of the old (list of strings) syntax, since supporting both formats is a huge pain 😅

hairyhenderson:
I've found a simpler way to fix this, right in the `UnmarshalYAML` function... commit coming in a few minutes...

hairyhenderson (approved):
Thanks again for your help with this @mihaitodor! 🙇‍♂️

PR #1581, opened 2022-12-18
Use one decoder for data.JSON, data.YAML. Deprecate data.JSONArray and data.YAMLArray

RaveNoX:
At first, thank you for very useful product.
I found it during searching for golang templating tool for DevOps needs after I started prototyping own tool with similar functionality, but now I decided to use gomplate and want to contribute.

During exploring and using gomplate, I faced with separation on converting on JSON and YAML data.
This separation was confused me, because from my point of view there's no need to have it.
So I decided to contribute some chages to make user experience smoother.

What actually changed in this PR:
- data.JSON and data.YAML is uses one unmarshaler (i.e. it now actually aliases) because valid JSON is valid YAML;
- both data.JSON and data.YAML supports convertion of any valid data types for this formats, including arrays;
- data.JSONArray and data.YAMLArray is deprecated, because this functionality can be achieved by using data.JSON and data.YAML, so it also only aliases, to keep compatibility;
- EJSON support is still here, but I added additinal stringifyMapKeys in output processing;

IMHO this changes will simplify usage of gomplate, because there's no need to decide which of YAML* or JSON* method use.

### DIFF
--- a/data/data.go
+++ b/data/data.go
@@ -64,6 +64,10 @@ func decryptEJSON(in string) (map[string]interface{}, error) {
 		return nil, errors.WithStack(err)
 	}
 	delete(obj, ejsonJson.PublicKeyField)
+
+	ret, _ := stringifyMapKeys(obj)
+	obj = ret.(map[string]interface{})
+
 	return obj, nil
 }
 
--- a/data/data.go
+++ b/data/data.go
@@ -8,7 +8,6 @@ import (
 	"bytes"
 	"encoding/csv"
 	"encoding/json"
-	"fmt"
 	"io"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


hairyhenderson:
Hi @RaveNoX,

Thanks for taking your time to work on this.

The JSON and YAML parsers were already using the same unmarshaler, with the difference that EJSON wasn't supported by the `data.YAML` function. Is there a reason you need that feature?

It is useful to have one function handling both objects and arrays, so that's a welcome change. However, the way you've done it breaks API compatibility with no deprecation (the `data.YAMLArray` function is part of stable API).

I would have appreciated if you had first filed an issue or started a [discussion](https://github.com/hairyhenderson/gomplate/discussions), before taking the time to issue this PR.

Before reviewing this further, I'd like to understand more about why you feel these changes are necessary - what specifically was confusing? What wasn't smooth about your user experience?

See https://github.com/hairyhenderson/gomplate/blob/main/CONTRIBUTING.md for instructions on how best to contribute with the least friction.

PR #1585, opened 2022-12-21
Add coll.JQ using gojq library

### LINKED ISSUE #1584: Feature Request: Add jq support
The [jq transformation language](https://stedolan.github.io/jq/manual/) using the [gojq library](https://github.com/itchyny/gojq) would be a great addition to gomplate to enable complex data processing that is often required.

**Proposed Usage:**
```
gomplate \
  -i '{{ .books | jq `[.works[]|{"title":.title,"authors":[.authors[].name],"published":.first_publish_year}]` }}' \
  -c books=https://openlibrary.org/subjects/fantasy.json
```

ahochsteger:
This PR adds support for the jq language to filter/transform complex data structures.
See [docs/content/functions/coll.md] for usage documentation.

Tasks:
- [X] Add `coll.JQ` implementation
- [X] Add documentation
- [x] Add tests

Questions:
- [X] Is it allowed for collection functions (`coll.JQ` in this case) to return `nil` if the expression evaluates to `nil` (or JSON `null`) or should an error be returned instead?
  -> Yes, should be ok. Should be mentioned in the documentation.
- [X] I'm not sure if it is a good idea to magically turn an array with one result into a single item as it is done for `jsonpath` and currently done for `jq` the same way. If an array returns just one element by accident it may break the template.
  -> Since there's no easy way to distinguish between an array with a single element and just a single element in the result it is kept the way it is done for `jsonpath`.

Future enhancements:
- Add variable binding (use gomplate variables in JQ expressions). It's not straight forward to implement in go templates.

Resolves #1584

### DIFF
--- a/coll/jq.go
+++ b/coll/jq.go
@@ -1,18 +1,20 @@
 package coll
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/itchyny/gojq"
 )
 
 // JQ -
-func JQ(jqExpr string, in interface{}) (interface{}, error) {
+func JQ(ctx context.Context, jqExpr string, in interface{}) (interface{}, error) {
 	query, err := gojq.Parse(jqExpr)
 	if err != nil {
 		return nil, fmt.Errorf("couldn't parse JQ %s: %w", jqExpr, err)
 	}
-	iter := query.Run(in)
+
+	iter := query.RunWithContext(ctx, in)
 	var out interface{}
 	a := []interface{}{}
 	for {
--- a/coll/jq.go
+++ b/coll/jq.go
@@ -20,7 +20,7 @@ func JQ(jqExpr string, in interface{}) (interface{}, error) {
 			break
 		}
 		if err, ok := v.(error); ok {
-			return nil, errors.Wrap(err, "executing JQ failed")
+			return nil, fmt.Errorf("executing JQ failed: %w", err)
 		}
 		if v != nil { // TODO: Check, if nil may be a valid result
 			a = append(a, v)
--- a/coll/jq.go
+++ b/coll/jq.go
@@ -1,0 +1,36 @@
+package coll
+
+import (
+	"github.com/itchyny/gojq"
+	"github.com/pkg/errors"
+)
+
+// JQ -
+func JQ(jqExpr string, in interface{}) (interface{}, error) {
+	query, err := gojq.Parse(jqExpr)
+	if err != nil {
+		return nil, errors.Wrapf(err, "couldn't parse JQ %s", jqExpr)
+	}
+	iter := query.Run(in)
+	var out interface{}
+	a := []interface{}{}
+	for {
+		v, ok := iter.Next()
+		if !ok {
+			break
+		}
+		if err, ok := v.(error); ok {
+			return nil, errors.Wrap(err, "executing JQ failed")
+		}
+		if v != nil {
+			a = append(a, v)
+		}
+	}
+	if len(a) == 1 {
+		out = a[0]
+	} else {
+		out = a
+	}
+
+	return out, nil
+}
--- a/coll/jq.go
+++ b/coll/jq.go
@@ -2,7 +2,9 @@ package coll
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"reflect"
 
 	"github.com/itchyny/gojq"
 )
@@ -11,7 +13,13 @@ import (
 func JQ(ctx context.Context, jqExpr string, in interface{}) (interface{}, error) {
 	query, err := gojq.Parse(jqExpr)
 	if err != nil {
-		return nil, fmt.Errorf("couldn't parse JQ %s: %w", jqExpr, err)
+		return nil, fmt.Errorf("jq parsing expression %q: %w", jqExpr, err)
+	}
+
+	// convert input to a supported type, if necessary
+	in, err = jqConvertType(in)
+	if err != nil {
+		return nil, fmt.Errorf("jq type conversion: %w", err)
 	}
 
 	iter := query.RunWithContext(ctx, in)
@@ -23,7 +31,7 @@ func JQ(ctx context.Context, jqExpr string, in interface{}) (interface{}, error)
 			break
 		}
 		if err, ok := v.(error); ok {
-			return nil, fmt.Errorf("executing JQ failed: %w", err)
+			return nil, fmt.Errorf("jq execution: %w", err)
 		}
 		a = append(a, v)
 	}
@@ -34,4 +42,55 @@ func JQ(ctx context.Context, jqExpr string, in interface{}) (interface{}, error)
 	}
 
 	return out, nil
+}
+
+// jqConvertType converts the input to a map[string]interface{}, []interface{},
+// or other supported primitive JSON types.
+func jqConvertType(in interface{}) (interface{}, error) {
+	// if it's already a supported type, pass it through
+	switch in.(type) {
+	case map[string]interface{}, []interface{},
+		string, []byte,
+		nil, bool,
+		int, int8, int16, int32, int64,
+		uint, uint8, uint16, uint32, uint64,
+		float32, float64:
+		return in, nil
+	}
+
+	inType := reflect.TypeOf(in)
+	value := reflect.ValueOf(in)
+
+	// pointers need to be dereferenced first
+	if inType.Kind() == reflect.Ptr {
+		inType = inType.Elem()
+		value = value.Elem()
+	}
+
+	mapType := reflect.TypeOf(map[string]interface{}{})
+	sliceType := reflect.TypeOf([]interface{}{})
+	// if it can be converted to a map or slice, do that
+	if inType.ConvertibleTo(mapType) {
+		return value.Convert(mapType).Interface(), nil
+	} else if inType.ConvertibleTo(sliceType) {
+		return value.Convert(sliceType).Interface(), nil
+	}
+
+	// if it's a struct, the simplest (though not necessarily most efficient)
+	// is to JSON marshal/unmarshal it
+	if inType.Kind() == reflect.Struct {
+		b, err := json.Marshal(in)
+		if err != nil {
+			return nil, fmt.Errorf("json marshal struct: %w", err)
+		}
+		var m map[string]interface{}
+		err = json.Unmarshal(b, &m)
+		if err != nil {
+			return nil, fmt.Errorf("json unmarshal struct: %w", err)
+		}
+		return m, nil
+	}
+
+	// we maybe don't need to convert the value, so return it as-is
+	return in, nil
 }
--- a/coll/jq.go
+++ b/coll/jq.go
@@ -22,7 +22,7 @@ func JQ(jqExpr string, in interface{}) (interface{}, error) {
 		if err, ok := v.(error); ok {
 			return nil, errors.Wrap(err, "executing JQ failed")
 		}
-		if v != nil {
+		if v != nil { // TODO: Check, if nil may be a valid result
 			a = append(a, v)
 		}
 	}
--- a/coll/jq.go
+++ b/coll/jq.go
@@ -9,7 +9,7 @@ import (
 func JQ(jqExpr string, in interface{}) (interface{}, error) {
 	query, err := gojq.Parse(jqExpr)
 	if err != nil {
-		return nil, errors.Wrapf(err, "couldn't parse JQ %s", jqExpr)
+		return nil, fmt.Errorf("couldn't parse JQ %s: %w", jqExpr, err)
 	}
 	iter := query.Run(in)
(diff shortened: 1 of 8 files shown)


ahochsteger:
@hairyhenderson, I'd like to allow variables defined in the template to be used within JQ expressions.

[gojq](https://github.com/itchyny/gojq#usage-as-a-library) provides a way to do that using [`gojq.WithVariables()`](https://pkg.go.dev/github.com/itchyny/gojq#WithVariables), but I'm unable to find a way to get the required information (names + values of variables).

This is how I'd use it (using hard-coded variables + values as an example):
```go
func JQ(jqExpr string, in interface{}) (interface{}, error) {
	query, err := gojq.Parse(jqExpr)
	variables := []string{"$testVariable1", "$testVariable2"} // TODO: Use variable names from context
	code, err := gojq.Compile(query, gojq.WithVariables(variables))
	iter := code.Run(in, "test value 1", "test value 2") // TODO: Use variable values from context
        ...
```
Is there a way to do what I want?

coll/jq.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"github.com/itchyny/gojq"
+	"github.com/pkg/errors"
+)
+
+// JQ -
+func JQ(jqExpr string, in interface{}) (interface{}, error) {
+	query, err := gojq.Parse(jqExpr)
+	if err != nil {
+		return nil, errors.Wrapf(err, "couldn't parse JQ %s", jqExpr)
```
hairyhenderson, line 12: I'd rather use `fmt.Errorf` than `pkg/errors` (which reminds me, I thought I got rid of uses of pkg/errors 🤔)

```suggestion
		return nil, fmt.Errorf("couldn't parse JQ %s: %w", jqExpr, err)
```

docs-src/content/functions/coll.yml
```diff
@@ -136,6 +136,37 @@ funcs:
       - |
         $ gomplate -i '{{ .books | jsonpath `$..works[?( @.edition_count > 400 )].title` }}' -c books=https://openlibrary.org/subjects/fantasy.json
         [Alice's Adventures in Wonderland Gulliver's Travels]
+  - name: coll.JQ
+    alias: jsonpath
```
hairyhenderson, line 140:
```suggestion
    alias: jq
```

docs-src/content/functions/coll.yml
```diff
@@ -136,6 +136,37 @@ funcs:
...
+    description: |
+      Filters an input object or list using the JQ language implemented by gojq (https://github.com/itchyny/gojq).
```
hairyhenderson, line 142:
```suggestion
      Filters an input object or list using the [jq](https://stedolan.github.io/jq/) language, as implemented by [gojq](https://github.com/itchyny/gojq).
```

docs/content/functions/coll.md
```diff
@@ -200,6 +200,51 @@ $ gomplate -i '{{ .books | jsonpath `$..works[?( @.edition_count > 400 )].title`
...
+
+Filters an input object or list using the JQ language implemented by gojq.
+
+Any object or list may be used as input. The output depends somewhat on the expression; if multiple items are matched, an array is returned.
+
+JQ filter expressions can be tested at https://jqplay.org/
+
+[jq Manual]: https://stedolan.github.io/jq/manual/
+[gojq library]: https://github.com/itchyny/gojq
+[gojq differences to jq]: https://github.com/itchyny/gojq#difference-to-jq
```
hairyhenderson, line 215: these look different from the input - can you re-generate with `make docs/content/functions/coll.md` (after addressing other comments in the doc source)?

docs-src/content/functions/coll.yml
```diff
@@ -136,6 +136,37 @@ funcs:
...
+
+      Any object or list may be used as input. The output depends somewhat on the expression; if multiple items are matched, an array is returned.
+
+      JQ filter expressions can be tested at https://jqplay.org/
+
+      [jq Manual]: https://stedolan.github.io/jq/manual/
+      [gojq differences to jq]: https://github.com/itchyny/gojq#difference-to-jq
```
hairyhenderson, line 149: these markdown footers aren't referenced anywhere, so are not useful. Maybe something like:

```suggestion
      See also:
      - [jq manual](https://stedolan.github.io/jq/manual/)
      - [gojq differences to jq](https://github.com/itchyny/gojq#difference-to-jq)
```

(See also the preview at https://deploy-preview-1585--gomplate-docs.netlify.app/functions/coll/#coll-jq)

docs-src/content/functions/coll.yml
```diff
@@ -136,6 +136,37 @@ funcs:
       - |
         $ gomplate -i '{{ .books | jsonpath `$..works[?( @.edition_count > 400 )].title` }}' -c books=https://openlibrary.org/subjects/fantasy.json
         [Alice's Adventures in Wonderland Gulliver's Travels]
+  - name: coll.JQ
+    alias: jsonpath
+    description: |
+      Filters an input object or list using the JQ language implemented by gojq (https://github.com/itchyny/gojq).
+
+      Any object or list may be used as input. The output depends somewhat on the expression; if multiple items are matched, an array is returned.
```
hairyhenderson, line 144: Could a string be used as input too? If not that should be pointed out, and a workaround suggested (i.e. parse with `data.JSON` first, etc)

coll/jq_test.go
```diff
@@ -0,0 +1,150 @@
...
+			Color: "red",
+		},
+		safe: "hidden",
+	}
+
+	// TODO: Check if this is a valid test case (taken from jsonpath_test.go) since the struct
+	// had to be converted to JSON and parsed from it again to be able to process using gojq.
+	v := map[string]interface{}{}
+	b, err := json.Marshal(structIn)
+	err = json.Unmarshal(b, &v)
```
hairyhenderson, line 140:
```suggestion
	b, err := json.Marshal(structIn)
	assert.NoError(err)
	err = json.Unmarshal(b, &v)
	assert.NoError(err)
```

coll/jq.go
```diff
@@ -0,0 +1,36 @@
...
+	iter := query.Run(in)
+	var out interface{}
+	a := []interface{}{}
+	for {
+		v, ok := iter.Next()
+		if !ok {
+			break
+		}
+		if err, ok := v.(error); ok {
+			return nil, errors.Wrap(err, "executing JQ failed")
```
hairyhenderson, line 23:
```suggestion
			return nil, fmt.Errorf("executing JQ failed: %w", err)
```

docs-src/content/functions/coll.yml
ahochsteger: Yes, any JSON datatype can be used as input, e.g. it works also for string or bolean as well:
```bash
$ gomplate -i '{{ "test" | jq `.` }}'
test
$ gomplate -i '{{ true | jq `.` }}'
true
```
I'll update the description as well.

hairyhenderson:
Thanks for this, @ahochsteger - as to your questions:

> Add variable binding (use gomplate variables in JQ expressions)

I'm not sure if this is going to be straightforward. Template functions don't have access to the template's context other than through passing in arguments. To allow this would probably open up security risks in the general case.

Perhaps an approach similar to the `template` keyword (like [`tmpl.Exec`](https://deploy-preview-1585--gomplate-docs.netlify.app/functions/tmpl/#tmpl-exec)) could be used, where a context value (often a map) is passed as an argument. For `coll.JQ`, it could look like this:

```
...
{{ "some input" | coll.JQ "some expression" (dict "a" "foo") }}
```

In that case a variable `$a` could be defined, with a value of `"foo"`.

However, it also may be a good idea to simply defer this and not support variable binding to begin with. It can be added later if there's demand.

>  Is it allowed for collection functions (coll.JQ in this case) to return nil if the expression evaluates to nil (or JSON null) or should an error be returned instead?

I don't know that I see a reason to not allow this, especially as it's certainly possible to have an explicit JSON value of `null`:

```console
$ echo '[null]' | jq '.[0]'
null
```

What's unfortunate about jq (and most JSON parsers) is there's effectively no difference between `{"a": null}` and `{}`. If I query `.a` for each of these inputs the result is always `null`.

Either way, I think it's OK to return `nil` in these sorts of cases. Just be sure to document it.

docs-src/content/functions/coll.yml
hairyhenderson: Ah - of course. It's probably also worth documenting the fact that string inputs are not first parsed as JSON then!

ahochsteger:
> What's unfortunate about jq (and most JSON parsers) is there's effectively no difference between `{"a": null}` and `{}`. If I query `.a` for each of these inputs the result is always `null`.

For this use case `jq` allows to check if a key exists using the `has` function:
```bash
$ gojq -n '{"a":null} | has("a")' 
true
$ gojq -n '{"a":null} | has("b")' 
false
```

coll/jq.go
```diff
@@ -0,0 +1,36 @@
...
+		}
+		if v != nil { // TODO: Check, if nil may be a valid result
+			a = append(a, v)
+		}
+	}
+	if len(a) == 1 {
+		out = a[0]
+	} else {
+		out = a
+	}
```
ahochsteger, line 33: @hairyhenderson I'm not sure if it is a good idea to magically turn an array with one result into a single item as it is done for `jsonpath` and (since it was my template for jq) is currently also done for jq the same way. If an array returns just one element by accident it may break the template.
Or do I miss something?

ahochsteger: It seems, that there's not an easy way to distinguish between an array with a single element and just an element in the result. Thus I keep it the same way as it is done for `jsonpath`.

coll/jq_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	structIn := &storeType{
+		Bicycle: &bicycleType{
+			Color: "red",
+		},
+		safe: "hidden",
+	}
+
+	// TODO: Check if this is a valid test case (taken from jsonpath_test.go) since the struct
+	// had to be converted to JSON and parsed from it again to be able to process using gojq.
```
hairyhenderson, line 137: 🤔

I suppose this is because JQ can't be used with arbitrary structs or other types, which is perhaps something we should allow...

In the same vein, it might be useful to be able to pass the template context ([`gomplate.tmplctx`](https://github.com/hairyhenderson/gomplate/blob/main/context.go#L12)) as the input for something like:

```console
$ gomplate -c books=https://openlibrary.org/subjects/fantasy.json -i '{{ jq `.works[].title` . }}'
```

Right now this fails with:

```
executing JQ failed: %!v(PANIC=Error method: invalid type: *gomplate.tmplctx (&map[books:map[key:/subjects/fantasy...
```

What's interesting in that particular case is that `gomplate.tmplctx` is a `map[string]interface{}`, which is in theory OK.

I'm going to hack on this a bit and make a commit if I can figure It out...

coll/jq.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/itchyny/gojq"
+)
+
+// JQ -
+func JQ(jqExpr string, in interface{}) (interface{}, error) {
+	query, err := gojq.Parse(jqExpr)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't parse JQ %s: %w", jqExpr, err)
+	}
+	iter := query.Run(in)
```
hairyhenderson, line 15: Just checking the docs for gojq it looks like there's a `query.RunWithContext` - I'd rather use this so we can support cancellation wherever possible.

I've made this change locally and will push a commit to your branch.

coll/jq_test.go
hairyhenderson: Ok, I've gotten this to work:

```console
$ bin/gomplate -c books=https://openlibrary.org/subjects/fantasy.json \
-i '{{ range jq `.books.works[].title` . -}}
title: {{.}}
{{end}}'
title: Alice's Adventures in Wonderland
title: Gulliver's Travels
title: Treasure Island
title: Through the Looking-Glass
title: A Midsummer Night's Dream
title: Il principe
title: The Wonderful Wizard of Oz
title: Avventure di Pinocchio
title: Alice's Adventures in Wonderland / Through the Looking Glass
title: Five Children and It
title: The Hobbit
title: Harry Potter and the Philosopher's Stone
```

Note that I've referenced the `books` datasource in the `.` context passed through as an argument.

hairyhenderson (approved):
Thanks again for working on this, @ahochsteger!

hairyhenderson:
I've added an integration test, and once the CI passes, this'll merge.

ahochsteger:
Thanks for merging it, @hairyhenderson!
Any plans for a new release containing this PR?

hairyhenderson:
@ahochsteger it'll be in 4.0.0 - too much has changed in `main` for it to be straightforward to go into a new 3.x release unfortunately 😞

I have no target date for 4.0, but in the meantime you can build your own binary from `main` to use this...

PR #1607, opened 2023-01-25
Add support for vault aws iam auth

cludden:
This change introduces support for the iam flavor of vault's aws auth engine. While the referenced, unmerged PR introduces a similar feature set, this implementation opts for using the same go libraries used by the vault cli for generating the login parameters.

related-pr: https://github.com/hairyhenderson/gomplate/pull/191


hairyhenderson:
@cludden thanks for working on this, however it would've been ideal if you'd opened an issue first...

In gomplate v4, which will hopefully be the next release, I'm moving over to use [go-fsimpl](https://github.com/hairyhenderson/go-fsimpl) for all datasources, and the `vaultfs` implementation already supports this: https://pkg.go.dev/github.com/hairyhenderson/go-fsimpl/vaultfs/vaultauth

I've been (slowly) working on incorporating that in #1336.

cludden:
I normally would have, but saw there was an open, but very out-of-date PR for this very feature, so referenced that in my commit instead. Happy to close this in favor of the upcoming v4 implementation..

hairyhenderson:
@cludden ok, fair enough! thanks for understanding 🙂

PR #1690, opened 2023-03-29
fix reading templates in `gomplate.RunTemplates()`

thesuperzapper:
Currently, when a downstream go package imports `github.com/hairyhenderson/gomplate/v3` and invokes `gomplate.RunTemplates()` with a non-empty `Templates` list in its `gomplate.Config`, it will always get a nil pointer dereference error.

For example, here is some example code that fails before this PR, but succeeds after:

```go
import (
	gomplate "github.com/hairyhenderson/gomplate/v3"
)

gomplateConfig := &gomplate.Config{
	InputDir:  "MY_INPUT",
	OutputDir: "MY_OUTPUT",
	LDelim:    "{{",
	RDelim:    "}}",
	Contexts:  []string{"Values=values.yaml"},
	Templates: []string{"helpers=MY_HELPERS"},
}
err := gomplate.RunTemplates(gomplateConfig)
if err != nil {
	return err
}
```

---

This is because the `gomplate.RunTemplates()` only uses `context.Background()` for its context:

https://github.com/hairyhenderson/gomplate/blob/f3ba5d5b35a6db83b21d079eabb15d2eae37c906/gomplate.go#L21-L27

Which is missing the `filefs.FS` provider that the main gomplate CLI uses:

https://github.com/hairyhenderson/gomplate/blob/f3ba5d5b35a6db83b21d079eabb15d2eae37c906/internal/cmd/main.go#L170-L175

----

This PR fixes the `gomplate.RunTemplates()` function by using extending the background context it initializes with `filefs.FS`, just like the main gomplate CLI already does.

### DIFF
--- a/gomplate.go
+++ b/gomplate.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/hairyhenderson/go-fsimpl/filefs"
 	"path/filepath"
 	"strings"
 	"text/template"
@@ -23,7 +24,16 @@ func RunTemplates(o *Config) error {
 	if err != nil {
 		return err
 	}
-	return Run(context.Background(), cfg)
+
+	ctx := context.Background()
+
+	// inject a default filesystem provider for file:// URLs
+	// required for reading template files
+	if FSProviderFromContext(ctx) == nil {
+		ctx = ContextWithFSProvider(ctx, filefs.FS)
+	}
+
+	return Run(ctx, cfg)
 }
 
 // Run all gomplate templates specified by the given configuration


hairyhenderson:
Hi @thesuperzapper, thanks for the PR.

Did you notice that `RunTemplates` is deprecated? I'd rather not make any changes to it, since it'll likely be removed in v4.0.

Have you tried to use the new [`Renderer`](https://pkg.go.dev/github.com/hairyhenderson/gomplate/v4#Renderer) type instead?

If you have issues with that, please file an issue so we can discuss before working on a PR.

For now, I'm going to close this since I don't want to update deprecated code.

hairyhenderson:
> can you clarify if/how you intend for people to import `gomplate` as a go module and invoke it programmatically?
>
> As far as I can see, `RunTemplates` is quite different from `Renderer`, as while `RunTemplates` is similar to the gomplate CLI, `Renderer` is not, as it can't take parameters like `--input-dir`.

If you want a CLI-like experience, [`Run`](https://pkg.go.dev/github.com/hairyhenderson/gomplate/v4#Run) will be the way.

You're right though - `Config` is still internal. I haven't yet had a chance to properly expose that. That's one of a number of things remaining before I can release v4.0.


> My dream is to have an easy way to programmatically call an "embedded" version of gomplate in my go app, with access to the same arguments as the `gomplate` CLI itself.

I'd be interested in exactly what your use-case is. Which parts of gomplate are you interested in particularly? The template functions? The datasource support? Something else?

thesuperzapper:
> I'd be interested in exactly what your use-case is. Which parts of gomplate are you interested in particularly? The template functions? The datasource support? Something else?

@hairyhenderson my use-case is generating a folder of Kubernetes manifests (for an app called Kubeflow), based on a user-provided `values.yaml` file in a way that is similar to a helm chart. Kubeflow is too large for a single helm chart, so my app actually generates a large number of separate manifests (including some helm charts).

So effectively the functionality I need is to take an input folder (which contains gomplate templates in its files), and render the folder into a separate directory, which can then be applied to a users Kubernetes cluster.

My app is called `deployKF`, and the repos are here, but please note that the app is not quite ready for usage yet:
- https://github.com/deployKF/deployKF/tree/dev
- https://github.com/deployKF/cli/tree/dev

----

Either way, when you expose the Config class, you should also expose some helper functions to "gather" the templates like you can with the current `Run`.

hairyhenderson:
Thanks @thesuperzapper - that's some good context.

I took a bit of a deeper look at `Run` and the `Renderer` type, which reminded me why I did that back in #1422 (have a look at that and the linked issues for _history_).

I think my goal was to expose the minimum of what I thought was useful to folks who wanted to use gomplate programatically. I specifically excluded the `gatherTemplates` function as I didn't expect folks would need that functionality. It's honestly gotten far more convoluted than I ever intended, and if you're already writing code to call gomplate you can probably gather templates in a less complex way for your purposes.

From your use-case it sounds like gathering templates is fairly straightforward - just walk the input directory, include the files that make sense, and use that to create a `[]gomplate.Template` and feed that into [`RenderTemplates`](https://pkg.go.dev/github.com/hairyhenderson/gomplate/v4#Renderer.RenderTemplates).

Otherwise, I think my goal for v4 is to unexport the otherwise-unusable [`gomplate.Run`](https://pkg.go.dev/github.com/hairyhenderson/gomplate/v4#Run) function.

Does that make sense?

thesuperzapper:
@hairyhenderson while I am sure I can use the `RenderTemplates` function to emulate the behavior of the `gomplate` CLI itself, I am not sure I understand the concern about providing this functionality natively.

That is, why not expose an API which effectively lets users invoke the `gomplate` CLI from their go apps (for people who want something exactly like the `gomplate` CLI, rather than just wanting its rendering engine)?

hairyhenderson:
> why not expose an API which effectively lets users invoke the `gomplate` CLI from their go apps

I'd rather answer the question "why" than "why not"... Saying "no" to a feature request is a temporary condition - given new information and/or increased demand over time that can always change to a "yes".

The burden of support for an exported API is on me, and I'm not comfortable with the state of that API right now.

The other angle is - if you just want to use `gomplate` invoked from some Go app, why not just use [`os/exec`](https://pkg.go.dev/os/exec#CommandContext) to spawn it?

thesuperzapper:
@hairyhenderson first, I greatly appreciate your effort on gomplate, as I am intimately aware of the pain in being an open-source maintainer.

In response to your question about using `os/exec`, I am not really comfortable with this as it would require me to automatically download the correct gomplate binary into the user's environment (or package the required one in some messy way).

In any case, I hope we can find a low-maintenance way to expose the CLI-like functionality of gomplate as a go API, but until that point, I guess I will continue with my slightly modified fork.

hairyhenderson:
Thanks for understanding, @thesuperzapper!

Once v4 is out, please feel welcome to revisit this - there may be a way that feels "cleaner" by then.

PR #1691, opened 2023-03-29
expose SuppressEmpty in `gomplate.Config`

thesuperzapper:
This PR adds the currently missing `SuppressEmpty` config to the public `gomplate.Config` class, so people packaging gomplate into their apps can enable this functionality when using `gomplate.RunTemplates()`.

### DIFF
--- a/config.go
+++ b/config.go
@@ -36,6 +36,8 @@ type Config struct {
 	RDelim string
 
 	Templates []string
+
+	SuppressEmpty bool
 }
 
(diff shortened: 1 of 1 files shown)


hairyhenderson:
The `Config` type is deprecated (see [the docs](https://pkg.go.dev/github.com/hairyhenderson/gomplate/v4#Config)), so I won't accept changes to it.

Please file an issue before PRs to discuss changes like this!

PR #1713, opened 2023-04-11
Add azure.Meta function to query Azure IMDS endpoint

jkroepke:
Based on https://github.com/hairyhenderson/gomplate/pull/891

Added `azure.Meta` similar to `gcp.Meta`.

Upstream: https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service

---

I tested the examples from docs on a real Azure Machine.

### DIFF
--- a/azure/meta.go
+++ b/azure/meta.go
@@ -12,9 +12,9 @@ import (
 	"github.com/hairyhenderson/gomplate/v4/env"
 )
 
-// DefaultEndpoint is the IMDS endpoint at Azure
+// defaultEndpoint is the IMDS endpoint at Azure
 // https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service
-var DefaultEndpoint = "http://169.254.169.254"
+const defaultEndpoint = "http://169.254.169.254"
 
 var (
 	// co is a ClientOptions populated from the environment.
@@ -29,7 +29,6 @@ type ClientOptions struct {
 }
 
 // GetClientOptions - Centralised reading of AZURE_TIMEOUT
-// ... but cannot use in vault/auth.go as different strconv.Atoi error handling
 func GetClientOptions() ClientOptions {
 	coInit.Do(func() {
 		timeout := env.Getenv("AZURE_TIMEOUT")
@@ -74,7 +73,7 @@ func NewMetaClient(options ClientOptions) *MetaClient {
 
 // Meta retrieves a value from the Azure Instance Metadata Service, returning the given default
 // if the service is unavailable or the requested URL does not exist.
-func (c *MetaClient) Meta(key, format, apiVersion string, def ...string) (string, error) {
+func (c *MetaClient) Meta(ctx context.Context, key, format, apiVersion string, def ...string) (string, error) {
 	url := c.endpoint + "/metadata/instance/" + key + "?api-version=" + apiVersion + "&format=" + format
 	return c.retrieveMetadata(url, def...)
 }
@@ -94,7 +93,7 @@ func (c *MetaClient) retrieveMetadata(url string, def ...string) (string, error)
 		c.client = &http.Client{Timeout: timeout, Transport: &http.Transport{Proxy: nil}}
 	}
 
-	request, err := http.NewRequest(http.MethodGet, url, nil)
+	request, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
 	if err != nil {
 		return returnDefault(def), nil
 	}
--- a/azure/meta.go
+++ b/azure/meta.go
@@ -74,8 +74,8 @@ func NewMetaClient(options ClientOptions) *MetaClient {
 
 // Meta retrieves a value from the Azure Instance Metadata Service, returning the given default
 // if the service is unavailable or the requested URL does not exist.
-func (c *MetaClient) Meta(key, format string, def ...string) (string, error) {
-	url := c.endpoint + "/metadata/instance/" + key + "?api-version=2021-12-13&format=" + format
+func (c *MetaClient) Meta(key, format, apiVersion string, def ...string) (string, error) {
+	url := c.endpoint + "/metadata/instance/" + key + "?api-version=" + apiVersion + "&format=" + format
 	return c.retrieveMetadata(url, def...)
 }
 
--- a/azure/meta.go
+++ b/azure/meta.go
@@ -1,0 +1,131 @@
+package azure
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/hairyhenderson/gomplate/v4/env"
+)
+
+// DefaultEndpoint is the IMDS endpoint at Azure
+// https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service
+var DefaultEndpoint = "http://169.254.169.254"
+
+var (
+	// co is a ClientOptions populated from the environment.
+	co ClientOptions
+	// coInit ensures that `co` is only set once.
+	coInit sync.Once
+)
+
+// ClientOptions contains various user-specifiable options for a MetaClient.
+type ClientOptions struct {
+	Timeout time.Duration
+}
+
+// GetClientOptions - Centralised reading of AZURE_TIMEOUT
+// ... but cannot use in vault/auth.go as different strconv.Atoi error handling
+func GetClientOptions() ClientOptions {
+	coInit.Do(func() {
+		timeout := env.Getenv("AZURE_TIMEOUT")
+		if timeout == "" {
+			timeout = "500"
+		}
+
+		t, err := strconv.Atoi(timeout)
+		if err != nil {
+			panic(fmt.Errorf("invalid AZURE_TIMEOUT value '%s' - must be an integer: %w", timeout, err))
+		}
+
+		co.Timeout = time.Duration(t) * time.Millisecond
+	})
+	return co
+}
+
+// MetaClient is used to access metadata accessible via the Azure compute instance
+// metadata service version 1.
+type MetaClient struct {
+	client   *http.Client
+	cache    map[string]string
+	endpoint string
+	options  ClientOptions
+}
+
+// NewMetaClient constructs a new MetaClient with the given ClientOptions. If the environment
+// contains a variable named `AZURE_META_ENDPOINT`, the client will address that, if not the
+// value of `DefaultEndpoint` is used.
+func NewMetaClient(options ClientOptions) *MetaClient {
+	endpoint := env.Getenv("AZURE_META_ENDPOINT")
+	if endpoint == "" {
+		endpoint = DefaultEndpoint
+	}
+
+	return &MetaClient{
+		cache:    make(map[string]string),
+		endpoint: endpoint,
+		options:  options,
+	}
+}
+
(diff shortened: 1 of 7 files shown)


hairyhenderson:
Hi @jkroepke - thanks for this. In general I think this is OK. I'll find some time over the next few days to properly review this.

jkroepke:
Hi @hairyhenderson 

thanks for the info!

azure/meta.go
```diff
@@ -0,0 +1,131 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/hairyhenderson/gomplate/v4/env"
+)
+
+// DefaultEndpoint is the IMDS endpoint at Azure
+// https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service
+var DefaultEndpoint = "http://169.254.169.254"
```
hairyhenderson, line 17: I know this is copied from `gcp/meta.go`, but it doesn't need to be exported (in either case):

```suggestion
// defaultEndpoint is the IMDS endpoint at Azure
// https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service
var defaultEndpoint = "http://169.254.169.254"
```

azure/meta.go
```diff
@@ -0,0 +1,131 @@
...
+	coInit sync.Once
+)
+
+// ClientOptions contains various user-specifiable options for a MetaClient.
+type ClientOptions struct {
+	Timeout time.Duration
+}
+
+// GetClientOptions - Centralised reading of AZURE_TIMEOUT
+// ... but cannot use in vault/auth.go as different strconv.Atoi error handling
```
hairyhenderson, line 32: this comment makes no sense (there's no Azure support in Vault... is there?)

```suggestion
```

funcs/azure.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// AddAzureFuncs -
+//
+// Deprecated: use [CreateAzureFuncs] instead
+func AddAzureFuncs(f map[string]interface{}) {
+	for k, v := range CreateAzureFuncs(context.Background()) {
+		f[k] = v
+	}
+}
```
hairyhenderson, line 26: let's not add immediately-deprecated functions

```suggestion
```

azure/meta.go
```diff
@@ -0,0 +1,131 @@
...
+func GetClientOptions() ClientOptions {
+	coInit.Do(func() {
```
hairyhenderson, line 34: Can you figure out why there's a `sync.Once` involved here? I've been digging through past history, and the pattern seems to come from https://github.com/hairyhenderson/gomplate/pull/248, though I honestly can't really tell _why_...

I think I'd rather remove the `coInit`, and return a `(ClientOptions, error)`, and error on `strconv.Atoi` rather than panic.

azure/meta.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/hairyhenderson/gomplate/v4/env"
+)
+
+// DefaultEndpoint is the IMDS endpoint at Azure
+// https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service
+var DefaultEndpoint = "http://169.254.169.254"
+
+var (
+	// co is a ClientOptions populated from the environment.
+	co ClientOptions
```
hairyhenderson, line 21: I don't understand why this is global - I'm pretty sure it doesn't need to be

azure/meta.go
```diff
@@ -0,0 +1,131 @@
...
+	return &MetaClient{
+		cache:    make(map[string]string),
+		endpoint: endpoint,
+		options:  options,
+	}
+}
+
+// Meta retrieves a value from the Azure Instance Metadata Service, returning the given default
+// if the service is unavailable or the requested URL does not exist.
+func (c *MetaClient) Meta(key, format, apiVersion string, def ...string) (string, error) {
```
hairyhenderson, line 77: Let's change this to take a `context.Context`, and propagate it through to `retrieveMetadata`:

```suggestion
func (c *MetaClient) Meta(ctx context.Context, key, format, apiVersion string, def ...string) (string, error) {
```

azure/meta.go
```diff
@@ -0,0 +1,131 @@
...
+
+	if c.client == nil {
+		timeout := c.options.Timeout
+		if timeout == 0 {
+			timeout = 500 * time.Millisecond
+		}
+		c.client = &http.Client{Timeout: timeout, Transport: &http.Transport{Proxy: nil}}
+	}
+
+	request, err := http.NewRequest(http.MethodGet, url, nil)
```
hairyhenderson, line 97: once `ctx` is passed into `retrieveMetadata`:

```suggestion
	request, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```

hairyhenderson:
I'm realizing there's a lot of legacy here (and in the GCP version, and in the original AWS version too).

In particular, the `ClientOptions` type isn't actually necessary at all if we just propagate the context, and use that for timeout rather than setting a client-wide `Timeout` (which is only used because that predates the `context` package).

I'm going to make these modifications, and then I'd ask that you pull them and test them out in Azure to make sure it still works as you expect 😉

hairyhenderson:
🤔 also the ability to provide a default value doesn't really make sense now that the [`conv.Default`/`default`](https://docs.gomplate.ca/functions/conv/#conv-default) function exists.

I'd rather that be used instead of duplicating functionality.

funcs/azure.go
```diff
@@ -0,0 +1,79 @@
...
+
+	if len(args) >= 2 {
+		if args[1] == "json" || args[1] == "text" {
+			format = args[1]
+			def = args[2:]
+		}
+	}
+	if len(args) >= 3 {
+		if found, _ := regexp.MatchString(`^\d{4}-\d{2}-\d{2}$`, args[2]); found {
+			apiVersion = args[2]
```
hairyhenderson, line 66: I'm not sure providing the API version as an argument makes a lot of sense - this seems like something that users will very rarely need to set to anything other than the latest.

And if they do need to change it I'd expect they'd need to change it to the same version for all calls in the template.

How about an environment variable for this instead, like `AZURE_API_VERSION`?

azure/meta.go
```diff
@@ -0,0 +1,131 @@
...
+	if value, ok := c.cache[url]; ok {
+		return value, nil
+	}
+
+	if c.client == nil {
+		timeout := c.options.Timeout
+		if timeout == 0 {
+			timeout = 500 * time.Millisecond
+		}
+		c.client = &http.Client{Timeout: timeout, Transport: &http.Transport{Proxy: nil}}
```
hairyhenderson, line 93: Can you explain why the `Transport` is being set? Why is `http.DefaultTransport` not appropriate?

azure/meta.go
```diff
@@ -0,0 +1,130 @@
...
+			panic(fmt.Errorf("invalid AZURE_TIMEOUT value '%s' - must be an integer: %w", timeout, err))
+		}
+
+		co.Timeout = time.Duration(t) * time.Millisecond
+	})
+	return co
+}
+
+// MetaClient is used to access metadata accessible via the Azure compute instance
+// metadata service version 1.
```
hairyhenderson, line 50:
```suggestion
// metadata service.
```

certainly not version 1...

funcs/azure.go
jkroepke: I added the API version here to make this function more future proof. The current default is the lastest, but it could be changed in future. 

> AZURE_API_VERSION

There is no global API Version at Azure. Each service can have a own API Version while the extact API Version may not exists on a different API.

It would work, since there is on IMDS for now. But if someone add Azure Key Vault, the AZURE_API_VERSION cant be re-used there.

azure/meta.go
jkroepke: By design, IMDS should never go through an proxy. 

https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows#proxies

Otherwise, you will recivce the IMDS from the proxy (if its an Azue Virtual Machine)

hairyhenderson: Thanks for the explanation - let's get a comment there explaining why, otherwise it'll confuse me again

funcs/azure.go
hairyhenderson: fair - then `AZURE_IMDS_API_VERSION` maybe...

> But if someone add Azure Key Vault [...]

yup, but I'd think it would make way more sense to add that as a datasource.

Which brings up [this conversation](https://github.com/hairyhenderson/gomplate/pull/891#issuecomment-656899540) again... Does `azure.Meta` _need_ to be a function? could there be an `azure+meta://` datasource instead?

hairyhenderson:
bringing [this](https://github.com/hairyhenderson/gomplate/pull/1713#discussion_r1167635698) out into the main thread so it doesn't get lost...

I'm now remembering that I had [thought of](https://github.com/hairyhenderson/gomplate/pull/891#issuecomment-656945101) re-implementing `aws.Meta` and `gcs.Meta` as datasources instead of functions.

Given that gomplate's datasource functionality has largely been extracted to [`go-fsimpl`](https://github.com/hairyhenderson/go-fsimpl), what do you think of _not_ adding `azure.Meta`, and instead adding a virtual `azure+meta://` filesystem? In some ways it might be far simpler to interact with (though that depends on your usage)...

Even if an `azure.Meta` function remains, it may still make sense to implement it as a filesystem (in `go-fsimpl`) and use that under the covers - providing the ability to choose either approach.

jkroepke:
> I'm now remembering that I had [thought of](https://github.com/hairyhenderson/gomplate/pull/891#issuecomment-656945101) re-implementing `aws.Meta` and `gcs.Meta` as datasources instead of functions.

Implemented them as datasource may works well for the gomplate CLI. I initially raise this PR to have the function on the Grafana Agent, which [implements](https://grafana.com/docs/agent/latest/cookbook/dynamic-configuration/03_advanced_datasources/01_aws/) gomplate as library. 

After re-reading the gomplate docs, there is a function to create a datasource https://docs.gomplate.ca/functions/data/#definedatasource inside a template. If datasources can be create inside a template on demand, it might be a better solution to implement the metadata services as datasource.

funcs/azure.go
jkroepke: From my point of view, both is fine. At datasource could reduce the calls to the IMDS service too.

But from user perspective, there is no difference between

```
{{- defineDatasource "imds" "azure+meta://instance" }}
Hello {{ (ds "imds").resourceId }}
```

or

```
Hello {{ azure.meta "instance/resourceId" "text" }}
```

hairyhenderson:
> I initially raise this PR to have the function on the Grafana Agent, which [implements](https://grafana.com/docs/agent/latest/cookbook/dynamic-configuration/03_advanced_datasources/01_aws/) gomplate as library.

Ah - Grafana Agent can use datasources too: https://grafana.com/docs/agent/latest/cookbook/dynamic-configuration/02_templates/02_datasources/

> I may stop working here until there a clarified way to go.

thanks - that's probably best.

As a workaround, you can probably use an HTTP datasource to reach the IMDS.

hairyhenderson:
I'm marking this as Draft as some kind of indication that it's "on hold" for now 😉

jkroepke:
> As a workaround, you can probably use an HTTP datasource to reach the IMDS.

The IMDS requires custom HTTP headers. I guess I cant set them through datasouce.

hairyhenderson:
With pure gomplate, you can - see https://docs.gomplate.ca/datasources/#sending-http-headers.

PR #1740, opened 2023-05-19
Autodetect KV2

mihaiush:
Autodetect KV2 if envvar VAULT_KV_AUTODETECT is set (1, yes, on)

### DIFF
--- a/data/datasource_vault.go
+++ b/data/datasource_vault.go
@@ -12,10 +12,7 @@ import (
 	"github.com/rs/zerolog"
 )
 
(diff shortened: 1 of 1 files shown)


hairyhenderson:
Hi @mihaiush, thanks for taking the time to put together this contribution.

I don't think I can accept this as-is, since I'm in the process of migrating to `go-fsimpl` for all datasource reads (see #1336).

Please also see the conversation in https://github.com/hairyhenderson/gomplate/issues/862, and if you have an issue that won't be covered by that, please file a new one.

PR #1781, opened 2023-07-02
Initial basic support for CUE

### LINKED ISSUE #1601: please add cue as a native format
[CUE](https://cuelang.org) is an intelligent configuration language and is a Go package.
We have many CUE files which are very clean and simple to manage, as well as some gomplate templates.
It would be really great if we could import these files directly without needing a cue to json pre-processing step.
thank you for considering.

hairyhenderson:
Fixes #1601

This adds CUE support to gomplate as a native datasource format, as well as `data.CUE`/`cue` and `data.ToCUE`/`toCUE` functions to allow inline parsing and rendering.

CUE has support for [modules and packages](https://cuelang.org/docs/concepts/packages/) (ability to import data from other files), which doesn't work just yet. ~Key to this is providing a filename, which is awkward with the current `parseData` signature.~

~I'll keep this in draft until I'm happy with a solution for this (or until I decide to simply not support importing).~

At least to start, CUE support will be single-file only, with no support for importing local packages.

### DIFF
--- a/data/data.go
+++ b/data/data.go
@@ -12,6 +12,9 @@ import (
 	"io"
 	"strings"
 
+	"cuelang.org/go/cue"
+	"cuelang.org/go/cue/cuecontext"
+	"cuelang.org/go/cue/format"
 	"github.com/Shopify/ejson"
 	ejsonJson "github.com/Shopify/ejson/json"
 	"github.com/hairyhenderson/gomplate/v4/conv"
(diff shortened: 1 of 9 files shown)


hairyhenderson:
Update on importing packages from the local filesystem: I may just be missing something, but I just don't think it's going to be possible. Even with providing a filename and import path at parse-time, imports fail. 

So at least initially, CUE support will be partial and not support imports (importing builtins does work, though).

PR #1835, opened 2023-09-11
Make `strings.Title` not lowercase acronyms

zregvart:
Sets the `NoLower` option so that acronyms are not lowercased.


hairyhenderson (approved):
Thanks @zregvart - this seems reasonable!

PR #1873, opened 2023-10-06
Explain that `index` supports nested keys

isavcic:
Hello, just had a scenario where I needed to access a nested YAML in which some key names should've been fetched from an environment variable. Hopefully this explains how to combine multiple nested keys with some additional functions if needed.


hairyhenderson (approved):
Thanks @isavcic, I appreciate this!

PR #1879, opened 2023-10-09
Add strings.Join function

ItalyPaleAle:
Example:

```
{{ strings.Join (coll.Slice 1 2 3) "," }}
```

returns "1,2,3"


hairyhenderson:
This already exists - see https://docs.gomplate.ca/functions/conv/#convjoin

PR #1885, opened 2023-10-19
docs: format user template in hello example

tenstad:
Correctly formats the example and makes it equal to the other one: https://github.com/hairyhenderson/gomplate/blob/d4d0df967d5ced07acde0c1e239da413b661f7bb/docs/content/_index.md#L24

### DIFF
--- a/docs/content/usage.md
+++ b/docs/content/usage.md
@@ -13,7 +13,7 @@ The template is read from standard in, and written to standard out.
 Use it like this:
 
 ```console
-$ echo "Hello, {{.Env.USER}}" | gomplate
+$ echo "Hello, {{ .Env.USER }}" | gomplate
 Hello, hairyhenderson
 ```
 


hairyhenderson:
> Correctly formats the example and makes it equal

can you clarify what's more "correct" about the new format? and, what do you mean by "equal"?

tenstad:
> > Correctly formats the example and makes it equal
> 
> can you clarify what's more "correct" about the new format? and, what do you mean by "equal"?

It is now equal to the other (referenced) _hello user_ example. And I would say it is "more correct" in that it now follows the convention of having padding within the brackets, which is in line with the rest of the documentation.

hairyhenderson:
Ah - apologies, I didn't realize the reference was from a separate file! Thanks for this.

hairyhenderson (approved):

PR #1900, opened 2023-11-07
Add support for Ed25519

horvski:


hairyhenderson:
Hi @horvski, thanks for working on this. I don't have time to review this in depth yet, but on the surface it looks like a reasonable new function.

Can I ask why you decided to name it `EDDSA` rather than something more specific like `Ed25519`? I think the more specific name would make more sense, given that EdDSA also has a variant based on edwards448.

Also in future please file an issue before issuing a PR (see [`CONTRIBUTING.md`](https://github.com/hairyhenderson/gomplate/blob/main/CONTRIBUTING.md)).

horvski:
Hi @hairyhenderson, no problem.  

> Can I ask why you decided to name it EDDSA rather than something more specific like Ed25519? I think the more specific name would make more sense, given that EdDSA also has a variant based on edwards448.

- I was going off of your ECDSA implementation, so that it would be ready somebody wanted to add Ed448 or Ed25519ph. For now though, I agree with you and renamed the function more specifically to Ed25519.

I will keep that in mind in the future, thanks.

horvski:
@hairyhenderson 

I also added signing and verifying with the key pair, to ensure that the encoding and decoding work as expected. This may be better off being split into separate unit tests. I figured this was good for now though, to at least know that the key pair was working.

crypto/eddsa.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Ed25519GenerateKey: failed to generate EDDSA private key: %w", err)
+	}
+	return pemEncodeEdPrivateKey(secret)
+}
+
+// Ed25519GenerateKeyFromSeed returns a PEM encoded Ed25519 Private Key from
+// `seed`. Panics if len(seed) is not SeedSize.
+func Ed25519GenerateKeyFromSeed(seed []byte) ([]byte, error) {
+	return pemEncodeEdPrivateKey(ed25519.NewKeyFromSeed(seed)) // Panics
```
hairyhenderson, line 24: This should not panic. Can you add a guard around `seed` to check its length first, and return an error instead?

crypto/eddsa.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// Ed25519DerivePublicKey returns an ed25519 Public Key from given PEM encoded
+// `privatekey`.
+func Ed25519DerivePublicKey(privatekey []byte) ([]byte, error) {
+	secret, err := ed25519DecodeFromPEM(privatekey)
+	if err != nil {
+		return nil, fmt.Errorf("EDDSADerivePublicKey: could not decode private key")
```
hairyhenderson, line 32: The error message should be prefixed with the function that was called, and should also wrap the error:

```suggestion
		return nil, fmt.Errorf("ed25519DecodeFromPEM: could not decode private key: %w", err)
```

crypto/eddsa.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	b, err := x509.MarshalPKIXPublicKey(secret.Public())
+	if err != nil {
+		return nil, fmt.Errorf("EDDSADerivePublicKey: failed to marshal PKIX public key: %w", err)
```
hairyhenderson, line 36:
```suggestion
		return nil, fmt.Errorf("MarshalPKIXPublicKey: failed to marshal PKIX public key: %w", err)
```

crypto/eddsa.go
```diff
@@ -0,0 +1,78 @@
...
+		Type:  "PUBLIC KEY",
+		Bytes: b,
+	}), nil
+}
+
+// pemEncodeEdPrivateKey is a convenience function for PEM encoding `secret`.
+func pemEncodeEdPrivateKey(secret ed25519.PrivateKey) ([]byte, error) {
+	der, err := x509.MarshalPKCS8PrivateKey(secret)
+	if err != nil {
+		return nil, fmt.Errorf("pemEncodeEdPrivateKey: failed to marshal ECDSA private key: %w", err)
```
hairyhenderson, line 48: This is ed25519, not ECDSA:

```suggestion
		return nil, fmt.Errorf("MarshalPKCS8PrivateKey: failed to marshal ed25519 private key: %w", err)
```

crypto/eddsa.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	block := &pem.Block{
+		Type:  "PRIVATE KEY",
+		Bytes: der,
+	}
+	buf := &bytes.Buffer{}
+	err = pem.Encode(buf, block)
+	if err != nil {
+		return nil, fmt.Errorf("pemEncodeEdPrivateKey: failed to encode generated EDDSA private key: pem encoding failed: %w", err)
```
hairyhenderson, line 57:
```suggestion
		return nil, fmt.Errorf("Encode: failed to encode generated ed25519 private key: PEM encoding failed: %w", err)
```

crypto/eddsa.go
```diff
@@ -0,0 +1,78 @@
...
+	if block == nil {
+		return nil, fmt.Errorf("ed25519DecodeFromPEM: failed to read key: no key found")
+	}
+	priv, err := x509.ParsePKCS8PrivateKey(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("ed25519DecodeFromPEM: invalid private key: %w", err)
+	}
+	secret, ok := priv.(ed25519.PrivateKey)
+	if !ok {
+		return nil, fmt.Errorf("ed25519DecodeFromPEM: invalid private key: %w", err)
```
hairyhenderson, line 75:
```suggestion
		return nil, fmt.Errorf("Decode: failed to read key: %w", err)
	}
	priv, err := x509.ParsePKCS8PrivateKey(block.Bytes)
	if err != nil {
		return nil, fmt.Errorf("ParsePKCS8PrivateKey: invalid private key: %w", err)
	}
	secret, ok := priv.(ed25519.PrivateKey)
	if !ok {
		return nil, fmt.Errorf("wrong type for private key (should be ed25519.PrivateKey, was %T", priv)
```

hairyhenderson (changes requested):
I've had a bit of a closer look, and made some review comments - thanks again for working on this @horvski!

In order to make this fully usable in gomplate, it'll need to also be added as a function (in `funcs/crypto.go`, etc...).

It'll also need documentation. I can handle those parts if you're not up for it, though it'll take me some time to get to it.

hairyhenderson:
BTW you'll also need to rebase for the CI to pass

crypto/eddsa.go
horvski: Added a guard.

crypto/eddsa.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, fmt.Errorf("MarshalPKCS8PrivateKey: failed to marshal ed25519 private key: %w", err)
+	}
+	block := &pem.Block{
+		Type:  "PRIVATE KEY",
+		Bytes: der,
+	}
+	buf := &bytes.Buffer{}
+	err = pem.Encode(buf, block)
+	if err != nil {
+		return nil, fmt.Errorf("Encode: failed to encode generated ed25519 private key: PEM encoding failed: %w", err)
```
hairyhenderson, line 60: why the linter cares about this and not the one on ln51, 🤷‍♂️

might as well make this a bit too - in particular the fact that it's a generated key isn't relevant in the scope of this function - it's also very unlikely for this to fail if everything that precedes it succeeds!

```suggestion
		return nil, fmt.Errorf("encode: PEM encoding: %w", err)
```

docs-src/content/functions/crypto.yml
```diff
@@ -158,6 +158,57 @@ funcs:
...
+      PEM-encoded PKCS#8 ASN.1 DER form.
+    pipeline: true
+    examples:
+      - |
+        $ gomplate -i '{{ crypto.Ed25519GenerateKey }}'
+        -----BEGIN PRIVATE KEY-----
+        ...
+  - name: crypto.Ed25519GenerateKeyFromSeed
+    experimental: true
+    released: v3.11.5
```
hairyhenderson, line 175:
```suggestion
    # released: v4.0.0
```

docs-src/content/functions/crypto.yml
```diff
@@ -158,6 +158,57 @@ funcs:
...
+      - name: seed
+        reqruired: true
+        description: the random seed encoded in either base64 or hex
+    examples:
+      - |
+        $ gomplate -i '{{ crypto.Ed25519GenerateKeyFromSeed base64 MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDA= }}'
+        -----BEGIN PRIVATE KEY-----
+        ...
+  - name: crypto.Ed25519DerivePublicKey
+    experimental: true
```
hairyhenderson, line 193:
```suggestion
    # released: v4.0.0
```

crypto/eddsa.go
hairyhenderson, line 1: can we rename this file as `ed25519.go`?

crypto/eddsa.go
```diff
@@ -0,0 +1,81 @@
...
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+)
+
+// Ed25519GenerateKey returns a random PEM encoded Ed25519 Private Key.
+func Ed25519GenerateKey() ([]byte, error) {
+	_, secret, err := ed25519.GenerateKey(rand.Reader)
+	if err != nil {
+		return nil, fmt.Errorf("Ed25519GenerateKey: failed to generate EDDSA private key: %w", err)
```
hairyhenderson, line 16: no need for the extra context:

```suggestion
		return nil, fmt.Errorf("generateKey: %w", err)
```

hairyhenderson:
Sorry for the delay - I've taken another pass and added some comments. If you rebase again the tests should pass!

horvski:
No worries! Sounds good, I'll do that now. I agree with your recent review, and yes, I think we should rename the file. I'll do that before I rebase.

docs-src/content/functions/crypto.yml
```diff
@@ -158,6 +158,57 @@ funcs:
         +hIz6+EUt/Db51awO7iCuRly5L4TZ+CnMAsIbtUOqsqwSQDtv0AclAuogmCst75o
         aztsmrD79OXXnhUlURI=
         -----END PUBLIC KEY-----
+  - name: crypto.Ed25519GenerateKey
+    experimental: true
+    released: v4.0.0
```
hairyhenderson, line 163: these need to be commented out (like in my original suggestion) beause this is not yet released (I'll update these when the release happens):

```suggestion
    # released: v4.0.0
```

docs-src/content/functions/crypto.yml
```diff
@@ -158,6 +158,57 @@ funcs:
...
+      PEM-encoded PKCS#8 ASN.1 DER form.
+    pipeline: true
+    examples:
+      - |
+        $ gomplate -i '{{ crypto.Ed25519GenerateKey }}'
+        -----BEGIN PRIVATE KEY-----
+        ...
+  - name: crypto.Ed25519GenerateKeyFromSeed
+    experimental: true
+    released: v4.0.0
```
hairyhenderson, line 175:
```suggestion
    # released: v4.0.0
```

hairyhenderson (commented):
Just one more minor change in the docs, otherwise this LGTM!

Can you also please regenerate the docs by running `make docs/content/functions/crypto.md` and committing the updated file?

docs-src/content/functions/crypto.yml
```diff
@@ -158,6 +158,57 @@ funcs:
...
+    arguments:
+      - name: encoding
+        required: true
+        description: the encoding that the seed is in (`hex` or `base64`)
+      - name: seed
+        reqruired: true
+        description: the random seed encoded in either base64 or hex
+    examples:
+      - |
+        $ gomplate -i '{{ crypto.Ed25519GenerateKeyFromSeed base64 MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDA= }}'
```
hairyhenderson, line 189:
```suggestion
        $ gomplate -i '{{ crypto.Ed25519GenerateKeyFromSeed "base64" MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDA= }}'
```

docs-src/content/functions/crypto.yml
horvski: I have it resolved locally, and am waiting to push up the latest changes. I sent a question to you via email.

horvski:
> Just one more minor change in the docs, otherwise this LGTM!
> 
> Can you also please regenerate the docs by running `make docs/content/functions/crypto.md` and committing the updated file?

I ran the command that you mentioned to create the new docs, but am getting this error:

```
$ make docs/content/functions/crypto.md
2023/11/30 10:36:00 Invalid Semantic Version
exit status 1
gomplate -d data=docs-src/content/functions/crypto.yml -f docs-src/content/functions/func_doc.md.tmpl -o docs/content/functions/crypto.md
10:36:08 ERR error="renderTemplate: failed to render template docs-src/content/functions/func_doc.md.tmpl: template: docs-src/content/functions/func_doc.md.tmpl:12:4: executing \"usage\" at <index $arguments $last>: error calling index: index out of range: -1"
make: *** [Makefile:214: docs/content/functions/crypto.md] Error 1
make: *** Deleting file 'docs/content/functions/crypto.md'
```

My guess is that since my fork does not have the semantic versioning, that's where the first error is coming from. The command seems to continue after this though, and I am not sure why it fails.

I got two different errors, but I can't get the first one again for some reason. At first, it was saying that crypto.md is already up to date. Now when I am running it, it keeps deleting the file and I have to git restore it.

Any clue of what I am doing wrong here?

docs-src/content/functions/crypto.yml
```diff
@@ -158,6 +158,57 @@ funcs:
         +hIz6+EUt/Db51awO7iCuRly5L4TZ+CnMAsIbtUOqsqwSQDtv0AclAuogmCst75o
         aztsmrD79OXXnhUlURI=
         -----END PUBLIC KEY-----
+  - name: crypto.Ed25519GenerateKey
+    experimental: true
+    released: v4.0.0
+    description: |
+      Generate a new Ed25519 Private Key and output in
+      PEM-encoded PKCS#8 ASN.1 DER form.
+    pipeline: true
```
hairyhenderson, line 167: because this function has no arguments, `pipeline` makes no sense (and breaks rendering with a cryptic error!)

```suggestion
```

hairyhenderson:
> I ran the command that you mentioned to create the new docs, but am getting this error:

I was able to duplicate this locally - it's because of `pipeline` - see https://github.com/hairyhenderson/gomplate/pull/1900#discussion_r1411185511

hairyhenderson:
> ```
> $ make docs/content/functions/crypto.md
> 2023/11/30 10:36:00 Invalid Semantic Version
> exit status 1
> ```

This can be safely ignored - it's from the Makefile, [here](https://github.com/hairyhenderson/gomplate/blob/main/Makefile#L21). It _might_ be happening because of how you forked the repo (maybe you didn't pull down tags? or maybe it's because you're working from your `main` branch?), but either way it's unimportant.

docs-src/content/functions/crypto.yml
hairyhenderson: Oh, I missed that - they both need to be in quotes as they're both strings 😉

docs-src/content/functions/crypto.yml
```diff
@@ -158,6 +158,56 @@ funcs:
...
+    arguments:
+      - name: encoding
+        required: true
+        description: the encoding that the seed is in (`hex` or `base64`)
+      - name: seed
+        required: true
+        description: the random seed encoded in either base64 or hex
+    examples:
+      - |
+        $ gomplate -i '{{ crypto.Ed25519GenerateKeyFromSeed "base64" MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDA= }}'
```
hairyhenderson, line 188: Sorry - I missed that the second value was also not in quotes 🤦‍♂️ 

```suggestion
        $ gomplate -i '{{ crypto.Ed25519GenerateKeyFromSeed "base64" "MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDA=" }}'
```

hairyhenderson:
@horvski just one final thing to fix, then a rebase, and it'll be good to go! Thanks for your patience and persistence on this 🙇‍♂️

hairyhenderson (approved):
Thanks @horvski! LGTM :shipit:

PR #1940, opened 2023-12-09
test: use `t.Setenv` to set env vars in tests

Juneezee:
Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

Reference: https://pkg.go.dev/testing#T.Setenv

```go
func TestFoo(t *testing.T) {
	// before
	os.Setenv(key, "new value")
	defer os.Unsetenv(key)
	
	// after
	t.Setenv(key, "new value")
}
```

This PR also splits large test cases into subtests. This is useful because VSCode is able to run subtest separately

image

### DIFF
--- a/aws/ec2info_test.go
+++ b/aws/ec2info_test.go
@@ -133,64 +133,64 @@ func TestNewEc2Info(t *testing.T) {
 }
 
 func TestGetRegion(t *testing.T) {
-	oldReg, ok := os.LookupEnv("AWS_REGION")
-	if ok {
-		defer os.Setenv("AWS_REGION", oldReg)
-	}
-	oldDefReg, ok := os.LookupEnv("AWS_DEFAULT_REGION")
-	if ok {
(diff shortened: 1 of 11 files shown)


hairyhenderson:
Hi @Juneezee, thanks for working on this. Most of these tests pre-date Go 1.17, which is why I didn't use `t.Setenv` (though you'll see that I have started using it in some newer tests in certain packages!).

I would have preferred to see the `t.Setenv` changes happen separately from the `t.Run` changes - as it is this is a fairly complex PR to review, and it will take some time, especially given that I want to make sure the test logic is entirely unchanged. But, I will review in depth when I have some time.

Juneezee:
Hi @hairyhenderson. No worries :smile:

hairyhenderson (approved):
Thanks @Juneezee! I've made some minor adjustments, otherwise this LGTM.

Issue #1948, opened 2023-12-19
[Feature] Handle with not existing keys


xxxcoltxxx:
Hello!
It would be nice to handle non-existing keys.
It will provide flexible control when working with non-existent keys and will make the template more readable.
It allows as to control where we should fallback to some default value and where we should fail template rendering.
Its easy to implement - just add command-line flag that controls the `missingkey=error|zero` option.

Current behavior:
```bash
$ gomplate -i 'Hi {{ .name }}'
Hi 14:06:57 ERR  error="failed to render template <arg>: template: <arg>:1:6: executing \"<arg>\" at <.name>: map has no entry for key \"name\""
$ gomplate -i 'Hi {{ if (has . "name") }}{{ .name }}{{ else }}Alex{{end}}'
Hi Alex
```

Proposed behavior:

```bash
$ gomplate --missing-key error -i 'Hi {{ .name }}'
Hi 14:06:57 ERR  error="failed to render template <arg>: template: <arg>:1:6: executing \"<arg>\" at <.name>: map has no entry for key \"name\""

$ gomplate --missing-key zero -i 'Hi {{ .name }}'
Hi <no value>

$ gomplate --missing-key zero -i 'Hi {{ .name | default "Alex" }}'
Hi Alex

$ gomplate --missingkey zero -i 'Hi {{ .name | required }}'
Hi 14:12:04 ERR  error="failed to render template <arg>: template: <arg>:1:11: executing \"<arg>\" at <required>: error calling required: can not render template: a required value was not set"
```

hairyhenderson:
Hi @xxxcoltxxx, thanks for filing this and thanks for the PR.

This has come up before, in #310 and #1362 at least, and I think in other issues as well.

I'm not sure why a flag didn't occur to me for controlling it, but at this point it seems reasonable.

gjrtimmer:
Is there any update on this? I hope this one makes it to a release soon, kinda need it.

hairyhenderson:
@gjrtimmer thanks for prodding - the PR is stuck, I will unstick it.

As for a release, this will be in v4, which won't be "soon" (depending on what "soon" means to you). In the meantime, there are workarounds (the [`test`](https://docs.gomplate.ca/functions/test/) namespace and [`coll.Has`](https://docs.gomplate.ca/functions/coll/#collhas)).

hairyhenderson:
Fixed by #1949

PR #1949, opened 2023-12-19
Add missing-key flag to manage behavior in case of non-existing key

xxxcoltxxx:
This pull request allow to control behavior in case of non-existent keys. It described in the issue #1948

### DIFF
--- a/docs/content/usage.md
+++ b/docs/content/usage.md
@@ -187,8 +187,9 @@ $ gomplate -c .=http://xkcd.com/info.0.json -i '<a href="{{ .img }}">{{ .title }
 Changes behavior if key not exist in the datasource.
 
 Available values:
-- `error` (default) - Exit with error if key not exist
-- `default`, `invalid` or `zero` - prints `<no value>` if key not exist
+- `error` (default): Execution stops immediately with an error.
+- `default` or `invalid`: Do nothing and continue execution. If printed, the result is the string `"<no value>"`.
+- `zero`: The operation returns the zero value for the element (which may be `nil`, in which case the string `"<no value>"` is printed).
 
 Examples:
 
--- a/docs/content/usage.md
+++ b/docs/content/usage.md
@@ -184,7 +184,7 @@ $ gomplate -c .=http://xkcd.com/info.0.json -i '<a href="{{ .img }}">{{ .title }
 
 ### `--missing-key`
 
-Changes behavior if key not exist in the datasource.
+Control the behavior during execution if a map is indexed with a key that is not present in the map
 
 Available values:
 - `error` (default): Execution stops immediately with an error.
--- a/docs/content/usage.md
+++ b/docs/content/usage.md
@@ -184,7 +184,7 @@ $ gomplate -c .=http://xkcd.com/info.0.json -i '<a href="{{ .img }}">{{ .title }
 
 ### `--missing-key`
 
-Control the behavior during execution if a map is indexed with a key that is not present in the map
+Control the behavior during execution if a map is indexed with a key that is not present in the map.
 
 Available values:
 - `error` (default): Execution stops immediately with an error.
--- a/docs/content/usage.md
+++ b/docs/content/usage.md
@@ -182,6 +182,37 @@ $ gomplate -c .=http://xkcd.com/info.0.json -i '<a href="{{ .img }}">{{ .title }
 <a href="https://imgs.xkcd.com/comics/diploma_legal_notes.png">Diploma Legal Notes</a>
 ```
 
+### `--missing-key`
+
+Changes behavior if key not exist in the datasource.
+
+Available values:
+- `error` (default) - Exit with error if key not exist
+- `default`, `invalid` or `zero` - prints `<no value>` if key not exist
+
+Examples:
+
+```console
+$ gomplate --missing-key error -i 'Hi {{ .name }}'
+Hi 14:06:57 ERR  error="failed to render template <arg>: template: <arg>:1:6: executing \"<arg>\" at <.name>: map has no entry for key \"name\""
+```
+
+```console
+$ gomplate --missing-key default -i 'Hi {{ .name }}'
+Hi <no value>
+```
+
+```console
+$ gomplate --missing-key zero -i 'Hi {{ .name | default "Alex" }}'
+Hi Alex
+```
+
+```console
+$ gomplate --missing-key zero -i 'Hi {{ .name | required }}'
+Hi 14:12:04 ERR  error="failed to render template <arg>: template: <arg>:1:11: executing \"<arg>\" at <required>: error calling required: can not render template: a required value was not set"
+```
+
+
 ### Overriding the template delimiters
 
 Sometimes it's necessary to override the default template delimiters (`{{`/`}}`).
--- a/gomplate_test.go
+++ b/gomplate_test.go
@@ -3,6 +3,7 @@ package gomplate
 import (
 	"bytes"
 	"context"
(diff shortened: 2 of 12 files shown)


gomplate_test.go
```diff
@@ -129,6 +130,38 @@ func TestHasTemplate(t *testing.T) {
 	assert.Equal(t, "bar", testTemplate(t, g, tmpl))
 }
 
+func TestMissingKey(t *testing.T) {
+	g := NewRenderer(Options{
+		MissingKey: "zero",
+	})
+	tmpl := `{{ .name }}`
+	assert.Equal(t, "<no value>", testTemplate(t, g, tmpl))
```
hairyhenderson, line 138: When it's `zero` shouldn't it be a zero value? from the docs, `<no value>` is the result of setting the `missingkey=default`

template.go
```diff
@@ -48,9 +48,9 @@ func copyFuncMap(funcMap template.FuncMap) template.FuncMap {
 }
 
 // parseTemplate - parses text as a Go template with the given name and options
-func parseTemplate(ctx context.Context, name, text string, funcs template.FuncMap, tmplctx interface{}, nested config.Templates, leftDelim, rightDelim string) (tmpl *template.Template, err error) {
+func parseTemplate(ctx context.Context, name, text string, funcs template.FuncMap, tmplctx interface{}, nested config.Templates, leftDelim, rightDelim string, missingKey string) (tmpl *template.Template, err error) {
 	tmpl = template.New(name)
-	tmpl.Option("missingkey=error")
+	tmpl.Option("missingkey=" + missingKey)
```
hairyhenderson, line 53: I think some validation of `missingKey` would be useful, throwing an error if the value is anything other than `error`, `zero`, or `default`.

hairyhenderson:
Thanks for working on this @xxxcoltxxx!

In addition to the comments I've made so far, I'd like to see an integration test added to cover both file-based config and flag-based config.

Also, the documentation needs updating as well.

gomplate_test.go
xxxcoltxxx: zero value for the interface{} is zero

`missionkey=default` prints `<no value>` always, but `missionkey=zero` prints `0` for the ints, for example

hairyhenderson: Right - that's why I'm confused about this test - why is `<no value>` being printed when `missingkey=zero`?

internal/config/configfile.go
```diff
@@ -465,6 +468,13 @@ func (c Config) Validate() (err error) {
 		}
 	}
 
+	if err == nil {
+		missingKeyValues := []string{"error", "zero"}
+		if !slices.Contains(missingKeyValues, c.MissingKey) {
+			err = fmt.Errorf("not allowed value for the 'missing-key' flag: %s. Allowed values: %s", c.MissingKey, strings.Join(missingKeyValues, ","))
```
hairyhenderson, line 474: the tests are failing currently because of this check - the value may also be empty, or `default`:

```suggestion
		missingKeyValues := []string{"", "error", "zero", "default"}
		if !slices.Contains(missingKeyValues, c.MissingKey) {
			err = fmt.Errorf("invalid value for 'missingKey': %q. Must be one of: %v", c.MissingKey, missingKeyValues))
```

template.go
hairyhenderson: I wrote this before I saw the validation already done in `configfile.go` 😅... However I think it's still necessary here, because this could get called programmatically (through `RenderTemplates`).

render.go
```diff
@@ -168,6 +173,7 @@ func NewRenderer(opts Options) *Renderer {
 		tctxAliases: tctxAliases,
 		lDelim:      opts.LDelim,
 		rDelim:      opts.RDelim,
+		missingKey:  opts.MissingKey,
```
hairyhenderson, line 176: this needs to default to `error` if it's empty, rather than sending an empty string

gomplate_test.go
xxxcoltxxx: because zero value for nil is `nil` and it prints `<no value>`

render.go
xxxcoltxxx: Done

internal/config/configfile.go
xxxcoltxxx: Done

xxxcoltxxx:
> Thanks for working on this @xxxcoltxxx!
> 
> In addition to the comments I've made so far, I'd like to see an integration test added to cover both file-based config and flag-based config.
> 
> Also, the documentation needs updating as well.

Added

template.go
xxxcoltxxx: I have added the additional check here

xxxcoltxxx:
> Thanks for working on this @xxxcoltxxx!
> 
> In addition to the comments I've made so far, I'd like to see an integration test added to cover both file-based config and flag-based config.
> 
> Also, the documentation needs updating as well.

Docs added

gomplate_test.go
hairyhenderson: ahhh, ok, thanks for explaining @xxxcoltxxx 🙂

docs/content/usage.md
```diff
@@ -182,6 +182,37 @@ $ gomplate -c .=http://xkcd.com/info.0.json -i '<a href="{{ .img }}">{{ .title }
 <a href="https://imgs.xkcd.com/comics/diploma_legal_notes.png">Diploma Legal Notes</a>
 ```
 
+### `--missing-key`
+
+Changes behavior if key not exist in the datasource.
```
hairyhenderson, line 187: this isn't just for datasources, so should be written to be more general, and I think a reference to https://pkg.go.dev/text/template#Template.Option is worthwhile

docs/content/usage.md
```diff
@@ -182,6 +182,37 @@ $ gomplate -c .=http://xkcd.com/info.0.json -i '<a href="{{ .img }}">{{ .title }
...
+
+Available values:
+- `error` (default) - Exit with error if key not exist
+- `default`, `invalid` or `zero` - prints `<no value>` if key not exist
```
hairyhenderson, line 191: Because there _might_ be situations where a non-`nil` zero value is present, I think this should be split up. And, we can take the wording from the Go docs too for consistency:

```suggestion
- `error` (default): Execution stops immediately with an error.
- `default` or `invalid`: Do nothing and continue execution. If printed, the result is the string `"<no value>"`.
- `zero`: The operation returns the zero value for the element (which may be `nil`, in which case the string `"<no value>"` is printed).
```

hairyhenderson:
Thanks @xxxcoltxxx - overall this looks good, just a few doc adjustments needed, as noted. The unit tests and linter need to be fixed, and then this can be merged.

Thanks again for your work on this!

xxxcoltxxx:
@hairyhenderson Fixed tests/linter/docs

Thanks for this repository. It helps a lot

hairyhenderson (approved):
Thanks again for working on this, @xxxcoltxxx!

philippeboyd:
Hi @hairyhenderson,

Would this feature be available in v3 (maybe `3.12`) or do we need to wait v4 or build `main` ourselves?

Thanks

hairyhenderson:
@philippeboyd there won't be a 3.12, you'll need to wait for 4.0 or build yourself. I've tagged a pre-release version (v4.0.0-pre-2) and will be building some binaries for it soon!

aviadhaham:
@hairyhenderson first of all, thank you for this great project.
I just wanted to ask when you think v4.0.0 will be released? I just found I need that feature 🥲

PR #1963, opened 2024-01-12
Add --exclude-processing option

### LINKED ISSUE #1911: Feature Request: Exclude from processing but still copy file to target folder
gomplate processes entire directory of files and sub-directories, and copies/creates/materializes each one in a desired target folder. The problem is when gomplate tries to process files like .jar or images and fails because of unexpected delimiter like characters in those files. This is totally not predictable because we cannot say what those binary files contain and if they happen to overlap with gomplate or any other templating engine syntax.

> It would be great if gomplate could support exclude such files from processing for templating but still copy them to the target folder. This is a classic case when the folder you are processing is a project template (like Java or other).

Exclusion may fall into one of the following categories

- `None` (default)
- `FromProcessing` (but shall be copied to the target folder)
- `FromCopying` (+Processing, the —exclude option)

An option like `—exclude-processing` that complements the existing `—exclude` would be great.

Manually copying using the `postexec` is possible but fragile because it is hard to guarantee that all the files not processed are picked up without skipping any file or folder and copied in the `postexec` step.

Ref: See [Discussion](https://github.com/hairyhenderson/gomplate/discussions/1884#discussioncomment-7537715) for details

lliissoonngg:
Fixes #1911 

This PR added a CLI option `--exclude-processing` and an equivelent config entry `excludeProcessing`, if it is set, the matching files will go with a different code path `copyFileToOutDir`.

To make the code reusable, the codes about file IO are offloaded from`fileToTemplate` into `readInFile`, `getOutfileHandler`, and are called by `copyFileToOutDir` as well. 

**How to test:**

Have a jpeg file in `in` directory.

**case 1: (test cli argument `--exclude-processing`)**

run:
`./gomplate --exclude-processing '*.jpg' --input-dir=in --output-dir=out`

the *.jpg exists in `out` directory  

**case 2: (test config file entry `excludeProcessing`)**

have a `.gomplate.yaml` file in the directory:
`
inputDir: in/
outputDir: out/
excludeProcessing:
  - '*.jpg'
 `

run:
`./gomplate --input-dir=in --output-dir=out`

the *.jpg exists in `out` directory

### DIFF
--- a/docs/content/usage.md
+++ b/docs/content/usage.md
@@ -130,6 +130,20 @@ $ gomplate --include *.tmpl --exclude foo*.tmpl --input-dir in/ --output-dir out
 
 This will cause only files ending in `.tmpl` to be processed, except for files with names beginning with `foo`: `template.tmpl` will be included, but `foo-template.tmpl` will not.
 
+### `--exclude-processing`
+
+When using the [`--input-dir`](#input-dir-and-output-dir) argument, it can be useful to skip some files from processing and copy them directly to the output directory. Like the `--exclude` flag, it takes a [`.gitignore`][]-style pattern, and any files match the pattern will be copied.
+
+_Note:_ These patterns are _not_ treated as filesystem globs, and so a pattern like `/foo/bar.json` will match relative to the input directory, not the root of the filesystem as they may appear!
+
+Examples:
+
+```console
+$ gomplate --exclude-processing `*.png` --input-dir in/ --output-dir out/
+```
+
+This will skip all `*.png` files in the `in/` directory from being processed, and copy them to the `out/` directory.
+
 #### `.gomplateignore` files
 
 You can also use a file named `.gomplateignore` containing one exclude pattern on each line. This has the same syntax as a [`.gitignore`][] file.
--- a/internal/cmd/main.go
+++ b/internal/cmd/main.go
@@ -138,7 +138,7 @@ func InitFlags(command *cobra.Command) {
 	command.Flags().StringP("in", "i", "", "Template `string` to process (alternative to --file and --input-dir)")
 	command.Flags().String("input-dir", "", "`directory` which is examined recursively for templates (alternative to --file and --in)")
 
-	command.Flags().StringSlice("exclude", []string{}, "glob of files to ignore")
+	command.Flags().StringSlice("exclude", []string{}, "glob of files to not parse")
 	command.Flags().StringSlice("exclude-processing", []string{}, "glob of files to be copied without parsing")
 	command.Flags().StringSlice("include", []string{}, "glob of files to parse")
 
--- a/internal/cmd/main.go
+++ b/internal/cmd/main.go
@@ -138,7 +138,8 @@ func InitFlags(command *cobra.Command) {
 	command.Flags().StringP("in", "i", "", "Template `string` to process (alternative to --file and --input-dir)")
 	command.Flags().String("input-dir", "", "`directory` which is examined recursively for templates (alternative to --file and --in)")
 
-	command.Flags().StringSlice("exclude", []string{}, "glob of files to not parse")
+	command.Flags().StringSlice("exclude", []string{}, "glob of files to ignore")
+	command.Flags().StringSlice("exclude-processing", []string{}, "glob of files to be copied without parsing")
 	command.Flags().StringSlice("include", []string{}, "glob of files to parse")
 
 	command.Flags().StringSliceP("out", "o", []string{"-"}, "output `file` name. Omit to use standard output.")
--- a/template.go
+++ b/template.go
@@ -204,7 +204,7 @@ func gatherTemplates(ctx context.Context, cfg *config.Config, outFileNamer func(
 		}}
 	case cfg.InputDir != "":
 		// input dirs presume output dirs are set too
-		templates, err = walkDir(ctx, cfg, cfg.InputDir, outFileNamer, cfg.ExcludeGlob, mode, modeOverride)
+		templates, err = walkDir(ctx, cfg, cfg.InputDir, outFileNamer, cfg.ExcludeGlob, cfg.ExcludeProcessingGlob, mode, modeOverride)
 		if err != nil {
 			return nil, fmt.Errorf("walkDir: %w", err)
 		}
@@ -224,7 +224,7 @@ func gatherTemplates(ctx context.Context, cfg *config.Config, outFileNamer func(
 // walkDir - given an input dir `dir` and an output dir `outDir`, and a list
 // of .gomplateignore and exclude globs (if any), walk the input directory and create a list of
 // tplate objects, and an error, if any.
-func walkDir(ctx context.Context, cfg *config.Config, dir string, outFileNamer func(context.Context, string) (string, error), excludeGlob []string, mode os.FileMode, modeOverride bool) ([]Template, error) {
+func walkDir(ctx context.Context, cfg *config.Config, dir string, outFileNamer func(context.Context, string) (string, error), excludeGlob []string, excludeProcessingGlob []string, mode os.FileMode, modeOverride bool) ([]Template, error) {
 	dir = filepath.ToSlash(filepath.Clean(dir))
 
 	// get a filesystem rooted in the same volume as dir (or / on non-Windows)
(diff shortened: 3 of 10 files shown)


hairyhenderson:
Hi @lliissoonngg, thanks for working on this.

Before I review this fully, a couple things:
1. some integration tests need to be written for this (see `internal/tests/integration/`)
2. I'm not sure about the name of this flag - "exclude-processing" isn't very clear - perhaps you have an idea on a better name? Or maybe @assayire from the original issue?

internal/cmd/main.go
```diff
@@ -139,7 +139,8 @@ func InitFlags(command *cobra.Command) {
 	command.Flags().StringP("in", "i", "", "Template `string` to process (alternative to --file and --input-dir)")
 	command.Flags().String("input-dir", "", "`directory` which is examined recursively for templates (alternative to --file and --in)")
 
-	command.Flags().StringSlice("exclude", []string{}, "glob of files to not parse")
+	command.Flags().StringSlice("exclude", []string{}, "glob of files to ignore")
```
hairyhenderson, line 142: I think the previous text was fine - "ignore" is somewhat ambiguous in this context

template.go
```diff
@@ -286,46 +312,84 @@ func walkDir(ctx context.Context, cfg *config.Config, dir string, outFileNamer f
 	return templates, nil
 }
 
-func fileToTemplate(ctx context.Context, cfg *config.Config, inFile, outFile string, mode os.FileMode, modeOverride bool) (Template, error) {
-	source := ""
+func readInFile(ctx context.Context, cfg *config.Config, inFile string, mode os.FileMode) (source string, newmode os.FileMode, err error) {
+	source = ""
+	newmode = mode
```
hairyhenderson, line 317: naked returns are an anti-pattern, so please do not use them.

lliissoonngg:
@assayire gently bump on this, could you take a look of the parameter naming in this PR?

assayire:
> ./gomplate --exclude-processing '*.jpg' --input-dir=in --output-dir=out

I am fine with the name.

Quick question: If I have exclude say jpg and png, would I do `--exclude-processing '*.jpg,*.png'`?

lliissoonngg:
> > ./gomplate --exclude-processing '*.jpg' --input-dir=in --output-dir=out
> 
> I am fine with the name.
> 
> Quick question: If I have exclude say jpg and png, would I do `--exclude-processing '*.jpg,*.png'`?

No, `--excludes` has higher priority, in this case jpg and png will not be copied. Let me know if this is good enough. I will update the doc after we all agree upon a working logic.

assayire:
> > > ./gomplate --exclude-processing '*.jpg' --input-dir=in --output-dir=out
> > 
> > 
> > I am fine with the name.
> > Quick question: If I have exclude say jpg and png, would I do `--exclude-processing '*.jpg,*.png'`?
> 
> No, `--excludes` has higher priority, in this case jpg and png will not be copied. Let me know if this is good enough. I will update the doc after we all agree upon a working logic.

Oops .. i think i typed it wrong. I meant to ask that if i have exclude multiple files from processing, would this work - `--exclude-processing '*.jpg,*.png,*.gif'`?

lliissoonngg:
> > > > ./gomplate --exclude-processing '*.jpg' --input-dir=in --output-dir=out
> > > 
> > > 
> > > I am fine with the name.
> > > Quick question: If I have exclude say jpg and png, would I do `--exclude-processing '*.jpg,*.png'`?
> > 
> > 
> > No, `--excludes` has higher priority, in this case jpg and png will not be copied. Let me know if this is good enough. I will update the doc after we all agree upon a working logic.
> 
> Oops .. i think i typed it wrong. I meant to ask that if i have exclude multiple files from processing, would this work - `--exclude-processing '*.jpg,*.png,*.gif'`?

Yes, multiple patterns works (exactly the same as `--excludes`).

hairyhenderson:
> would this work - `--exclude-processing '*.jpg,*.png,*.gif'`?

If it works like `--exclude` then I think it would not be a comma-separated string, but instead multiple options, so:

```
--exclude-processing '*.jpg' --exclude-processing '*.png' --exclude-processing '*.gif'
```

For something so verbose as this it would make more sense to use a config file, which would look like:

```yaml
excludeProcessing
  - '*.jpg'
  - '*.png'
  - '*.gif'
```

Hope that makes sense @assayire

hairyhenderson:
@lliissoonngg thanks for the updates. Can you please rebase? I've just merged #1336 which makes a large number of changes that will affect this.

I'll make another review pass after that.

tewfik-ghariani:
Would it be also possible to exclude the processing of certain files even if "input" and "output" are not specified ( regardless of the copying operation)?
For example
```
$ ls -1 helm-values/
template.yml.tmpl
values.yml

$ gomplate -f helm-values/ --exclude-processing  '*.yml'
```

hairyhenderson:
@tewfik-ghariani that command doesn't really make sense - `-f` must refers to a file, not a directory. This option would only make sense when using `--input-dir`.

assayire:
> > would this work - `--exclude-processing '*.jpg,*.png,*.gif'`?
> 
> If it works like `--exclude` then I think it would not be a comma-separated string, but instead multiple options, so:
> 
> ```
> --exclude-processing '*.jpg' --exclude-processing '*.png' --exclude-processing '*.gif'
> ```
> 
> For something so verbose as this it would make more sense to use a config file, which would look like:
> 
> ```yaml
> excludeProcessing
>   - '*.jpg'
>   - '*.png'
>   - '*.gif'
> ```
> 
> Hope that makes sense @assayire

@hairyhenderson  How would you specify it on command line to --exclude-processing? I mean the multiple files as you have specified in yaml.

hairyhenderson:
@assayire like I said:

> If it works like `--exclude` then I think it would not be a comma-separated string, but instead multiple options, so:
> ```
> --exclude-processing '*.jpg' --exclude-processing '*.png' --exclude-processing '*.gif'
> ```

docs/content/usage.md
```diff
@@ -130,6 +130,20 @@ $ gomplate --include *.tmpl --exclude foo*.tmpl --input-dir in/ --output-dir out
 
 This will cause only files ending in `.tmpl` to be processed, except for files with names beginning with `foo`: `template.tmpl` will be included, but `foo-template.tmpl` will not.
 
+### `--exclude-processing`
+
+When using the [`--input-dir`](#input-dir-and-output-dir) argument, it can be useful to skip some files from processing and copy them directly to the output directory. Like the `--exclude` flag, it takes a [`.gitignore`][]-style pattern, and any files match the pattern will be copied.
```
hairyhenderson, line 135:
```suggestion
When using the [`--input-dir`](#input-dir-and-output-dir) argument, it can be useful to skip some files from processing and copy them directly to the output directory. Like the `--exclude` flag, it takes a [`.gitignore`][]-style pattern, and any files matching the pattern will be copied.
```

assayire:
LGTM

template.go
```diff
@@ -265,8 +265,24 @@ func walkDir(ctx context.Context, cfg *config.Config, dir string, outFileNamer f
 		return nil, fmt.Errorf("ignore matching failed for %s: %w", dir, err)
 	}
 
+	excludeProcessingMatches, err := matcher.Matches(".", &xignore.MatchesOptions{
+		Ignorefile:    gomplateignore,
+		Nested:        true, // allow nested ignorefile
```
hairyhenderson, line 270: I think this means that the `.gomplateignore` file will be _also_ considered for `excludeProcessingMatches`, which doesn't make sense.

I'm not sure this will functionally cause a problem (because those files will already be completely ignored, and skipped), but it's liable to cause a bug down the road.

The `xignore` module that I'm using unfortunately doesn't seem to support _not_ specifying an `Ignorefile`. I think I'm going to need to replace that module soon.

For now, how about a comment:

```suggestion
		// TODO: fix or replace xignore module so we can avoid attempting to read the .gomplateignore file for both exclude and excludeProcessing patterns
		Ignorefile:    gomplateignore,
		Nested:        true, // allow nested ignorefile
```

hairyhenderson (approved):
Thanks again for this contribution @lliissoonngg! I'm going to rebase and merge this now, and will work (later) on replacing the `xignore` module.

mgbowman:
@assayire, it doesn't look like it. This was merged on Feb 6 and the [v3.11.7](https://github.com/hairyhenderson/gomplate/releases/tag/v3.11.7) release was published on Jan 23, which is a pity because I could _really_ use this for my current project. 😞 

@hairyhenderson any chance we can get a `v3.x` release with this option or do we need to wait for `v4.x`?

hairyhenderson:
This is a new feature and as such is going in the next release, which will be v4.

I could cut a v4.0.0-pre-3 pre-release if that helps.

(also in general please don't pose questions on closed issues, they're very easy to miss!)

PR #1976, opened 2024-01-26
Remove support for deprecated key/value array form of template config

hairyhenderson:


hairyhenderson:
actually I didn't want to do this right away, but later on in v4 instead...

PR #2024, opened 2024-03-18
Update syntax.md

cm-sentinel:
fix: wording was incorrect.

Details: Original text stated
```... newlines that come before the action) can be suppressed by placing a minus sign in front of the first set of delimiters (`{{`)...```

This appears to have been a typo. Following those directions would lead to (`-{{`). This would not behave as expected. What was originally intended was (`{{-`)

### DIFF
--- a/docs/content/syntax.md
+++ b/docs/content/syntax.md
@@ -51,7 +51,7 @@ Hello,  baz!
 
 This might not be desirable.
 
-You can use [Golang template syntax](https://golang.org/pkg/text/template/#hdr-Text_and_spaces) to fix this. Leading newlines (i.e. newlines that come before the action) can be suppressed by placing a minus sign in front of the first set of delimiters (`{{`). Putting the minus sign behind the trailing set of delimiters (`}}`) will suppress the newline _after_ the action. You can do both to suppress newlines entirely on that line.
+You can use [Golang template syntax](https://golang.org/pkg/text/template/#hdr-Text_and_spaces) to fix this. Leading newlines (i.e. newlines that come before the action) can be suppressed by placing a minus sign behind the first set of delimiters (`{{`). Putting the minus sign in front of the trailing set of delimiters (`}}`) will suppress the newline _after_ the action. You can do both to suppress newlines entirely on that line.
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/content/syntax.md [hunk 1]
```diff
@@ -51,7 +51,7 @@ Hello,  baz!
 
 This might not be desirable.
 
-You can use [Golang template syntax](https://golang.org/pkg/text/template/#hdr-Text_and_spaces) to fix this. Leading newlines (i.e. newlines that come before the action) can be suppressed by placing a minus sign in front of the first set of delimiters (`{{`). Putting the minus sign behind the trailing set of delimiters (`}}`) will suppress the newline _after_ the action. You can do both to suppress newlines entirely on that line.
+You can use [Golang template syntax](https://golang.org/pkg/text/template/#hdr-Text_and_spaces) to fix this. Leading newlines (i.e. newlines that come before the action) can be suppressed by placing a minus sign behind the first set of delimiters (`{{`). Putting the minus sign in front of the trailing set of delimiters (`}}`) will suppress the newline _after_ the action. You can do both to suppress newlines entirely on that line.
```
hairyhenderson, line 54: I guess it depends what you mean by "in front of" or "behind"... The original text was presuming "front" to be "to the right", and "behind" to be "to the left".

IMO neither of these are that clear.

The Go docs that describe this behaviour use better wording:

> [...] if an action's left delimiter (by default "{{") is followed immediately by a minus sign and white space, all trailing white space is trimmed from the immediately preceding text. Similarly, if the right delimiter ("}}") is preceded by white space and a minus sign [...]

So, how about:

```suggestion
You can use [Golang template syntax](https://golang.org/pkg/text/template/#hdr-Text_and_spaces) to fix this. Leading newlines (i.e. newlines that come before the action) can be suppressed by placing a minus sign and white space following the left delimiter (`{{`). Putting white space and a minus sign preceding the right delimiter (`}}`) will suppress the newline _after_ the action. You can do both to suppress newlines entirely on that line.
```

I'm still not super happy with how that reads though - perhaps we could simply copy the Go doc's text and insert it here in a way that makes sense.

PR #2059, opened 2024-05-05
Build(deps): replace inet.af/netaddr by github.com/inetaf/netaddr

vincentbernat:
inet.af domain was seized and is not functional anymore. Unfortunately, github.com/inetaf/netaddr is not a dropin replacement as its path has not been updated:

        module declares its path as: inet.af/netaddr
                but was required as: github.com/inetaf/netaddr

I don't know the best way to handle this. A `replace` would not help dependencies. Alternatively, support for `inet.af/netaddr` could be dropped as it is already broken due to the domain problem.

cc @bradfitz for insights (maybe push a last update to update the path?)

### DIFF
--- a/docs-src/content/functions/net.yml
+++ b/docs-src/content/functions/net.yml
@@ -7,7 +7,7 @@ preamble: |
 
   [RFC 4632]: http://tools.ietf.org/html/rfc4632
   [RFC 4291]: http://tools.ietf.org/html/rfc4291
-  [`inet.af/netaddr`]: https://pkg.go.dev/inet.af/netaddr
+  [`github.com/inetaf/netaddr`]: https://pkg.go.dev/github.com/inetaf/netaddr
   [`net`]: https://pkg.go.dev/net
 funcs:
   - name: net.LookupIP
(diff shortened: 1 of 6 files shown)


hairyhenderson:
_oof..._ that's rough. Thanks for the PR @vincentbernat - will look at it soon!

hairyhenderson:
hmm... I have some further thoughts on this... Given:

- https://github.com/inetaf/netaddr is archived
- there's very little in gomplate that still depends on `inet.af/netaddr`
- the functions that _do_ depend on it are all deprecated with replacements available in v4
- the go proxy has a cached version still (but maybe not forever?)

I think probably the best approach is to just leave things as-is for now, and work on removing support in the next major version (v5, because they're only just marked as deprecated as of v4).

Or are you running into issues now? (i.e. are you trying to build gomplate from source with `GOPROXY=direct`?)

vincentbernat:
You are right, I am using GOPROXY=direct, I don't remember why...

bradfitz:
That package says:

> Deprecated
> Please see https://pkg.go.dev/go4.org/netipx and the standard library's [net/netip](https://pkg.go.dev/net/netip)

(We moved it into the standard library a few releases ago.)

hairyhenderson:
Given the affected functions are deprecated in v4, and the module is more-or-less available with the module proxy, I'm going to close this and simply follow the original removal plan. If removal needs to be accelerated that's always an option too.

PR #2094, opened 2024-06-03
chore(api)!: Overhauling config and rendering types

### LINKED ISSUE #2073: Need to finalize gomplate.Renderer and related types before v4.0.0
It looks like the public interface for using `gomplete` is in flux as it is moved to version 4 so this is likely a "it's not quite done yet" question.  

The deprecated `RunTemplates` supports passing in `InputDir` and `OutputDir` which is super convenient when processing a directory since it builds the list of templates before calling the new `gomplate.Renderer#RenderTemplates`.  This appears to be currently broken on `gatherTemplates` when `cfg.InputDir != ""` since the `FSProvider` is unset on the context.  

Is the intent to have something similar to `RunTemplates` (or `Run`) with the new `Renderer` or should I enumerate the list of templates myself and provide it to `RenderTemplates`?

hairyhenderson:
Exporting `config.Config` and changing `Renderer`-related types as a result. A bunch of renames and shuffles. Everything should remain status quo from a CLI perspective, but this is one of the bigger breaking changes coming to v4.0 for programmatic users.

Fixes #2073


hairyhenderson:
> I need to decide whether I want the type in the root package and rename it to something other than `Config` so as not to conflict with the deprecated type, or go ahead and completely break things and remove `RunTemplates` and change the deprecated `Config` type to this new one instead.

I'm going to remove the deprecated `RunTemplates` and `Config` - they've been deprecated for 5 minor releases (since [3.7.0](https://github.com/hairyhenderson/gomplate/pull/853)), more than 4 years... It's time. But, I'll do that in a separate PR for clarity.

hairyhenderson:
#2095 has the removals, will rebase this after it merges

PR #2107, opened 2024-06-11
Fix example for Strings.Trim

krishnatippur:
Minor doc update: The current example is missing a trailing `'`.


hairyhenderson (approved):
Thank you @krishnatippur! 🙇‍♂️

Issue #2121, opened 2024-06-19
Tests failing on 32-bit builds due to unaligned struct in dependency


hairyhenderson:
This is blocking the Alpine updates (https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/67862) because the tests are failing on 32-bit builds (armhf, armv7, x86). I've been able to reproduce locally in a container (`docker run --platform=linux/armhf ...`)

The issue seems to be that the `keyvalue.Blob` type is misaligned:

```go
type Bytes struct {
	bytes  []byte
	length int64
	mu     *sync.Mutex
}
```

I think it should be:

```go
type Bytes struct {
	bytes  []byte
	length int64
	mu     *sync.Mutex
}
```

(at least that's what `fieldalignment` likes).

I'll fix this in a fork.

Here's the panic:

```
panic: unaligned 64-bit atomic operation [recovered]
        panic: unaligned 64-bit atomic operation

goroutine 44 [running]:
testing.tRunner.func1.2({0x1d34028, 0x292dba8})
        /usr/local/go/src/testing/testing.go:1631 +0x274
testing.tRunner.func1()
        /usr/local/go/src/testing/testing.go:1634 +0x3fc
panic({0x1d34028, 0x292dba8})
        /usr/local/go/src/runtime/panic.go:770 +0xfc
runtime/internal/atomic.panicUnaligned()
        /usr/local/go/src/runtime/internal/atomic/unaligned.go:8 +0x24
runtime/internal/atomic.Load64(0x405da1c)
        /usr/local/go/src/runtime/internal/atomic/atomic_arm.s:291 +0x14
github.com/hack-pad/hackpadfs/keyvalue/blob.(*Bytes).Len(0x405da10)
        /go/pkg/mod/github.com/hack-pad/hackpadfs@v0.2.1/keyvalue/blob/bytes.go:56 +0x24
github.com/hack-pad/hackpadfs/keyvalue.(*runOnceFileRecord).Size(0x405a758)
        /go/pkg/mod/github.com/hack-pad/hackpadfs@v0.2.1/keyvalue/record.go:159 +0xa8
github.com/hack-pad/hackpadfs/keyvalue.(*file).Truncate(0x4139070, 0x0)
        /go/pkg/mod/github.com/hack-pad/hackpadfs@v0.2.1/keyvalue/file.go:305 +0xf4
[...]

PR #2126, opened 2024-06-22
ci(test): Add 32-bit & ARM CI tests

hairyhenderson:
#2121 only popped up because the Alpine build runs tests on all platforms. I should at least run tests on 32-bit platforms in CI to make sure I don't hit errors like this again.

Also adding emulated tests (with QEMU) for ARM64 and ARMv7. They take a bit longer (because of the crypto differences), but it'll probably be worth it.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new build setup for 32-bit Linux platforms.
  - Added a dedicated Dockerfile (`Dockerfile.test32bit`) for 32-bit build environments.

- **Refactor**
  - Updated the existing Dockerfile to use a multi-stage build process and set environment variables for cross-platform compatibility.
  - Enhanced testing commands in the `Makefile` with race detection for supported platforms.

- **Chores**
  - Simplified the linting workflow by removing unnecessary arguments for a cleaner configuration.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,16 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/hairyhenderson/gomplate-ci-build
+    strategy:
+      fail-fast: false
+      matrix:
+        goarch: [ '386', 'amd64', 'arm', 'arm64' ]
+        include:
+          - goarch: 'arm'
+            goarm: '7'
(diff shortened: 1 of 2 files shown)


hairyhenderson:
@coderabbitai ignore

hairyhenderson:
Alright, I've got this to the point where it "properly" fails.

Pulling in my fork fix to see if that'll make it succeed.

PR #2132, opened 2024-06-24
ci(build): Add macOS tests

hairyhenderson:
Adds a matrix strategy to build & test on macOS.

## Summary by CodeRabbit

- **Chores**
  - Improved build job workflow by unifying Windows and macOS builds into a single job with a matrix strategy.
  - Dynamically set environment variables based on the operating system.
  - Added setup steps for macOS test dependencies.


hairyhenderson:
on second thought, I develop on macOS so this is actually fairly low-value. I'd have to shave a bunch more yaks to get the CI environment running properly and it's not really worth it.

PR #2202, opened 2024-08-05
Do not fail on path resolution for default config

nabokihms:
There is now an error after upgrading to the latest template with the config file support.
> Error: fsys for path .gomplate.yaml: resolve local path \".gomplate.yaml\": getwd: stat .: permission denied

It seems like the error happens during the local path resolution, which calls various os package methods, including getcwd.
If the config was not requested intentionally, I believe it is ok to skip these errors.

Context https://github.com/dexidp/dex/issues/3675

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -77,7 +77,10 @@ func readConfigFile(ctx context.Context, cmd *cobra.Command) (*gomplate.Config, 
 	// we only support loading configs from the local filesystem for now
 	fsys, err := datafs.FSysForPath(ctx, cfgFile)
 	if err != nil {
-		return nil, fmt.Errorf("fsys for path %v: %w", cfgFile, err)
+		if configRequired {
+			return nil, fmt.Errorf("fsys for path %v: %w", cfgFile, err)
+		}
+		return nil, nil
 	}
 
 	f, err := fsys.Open(cfgFile)


nabokihms:
Haven't managed to find a good workaround yet. Can we release a version with the fix anytime soon? Rolling back to previous gomplate version is always an option, but there are several security issues in v3.* with dependencies.

hairyhenderson:
Hi @nabokihms, thanks for this.

> upgrading to the latest template with the config file support.

hrm... I see you upgraded from 3.11.8 - it had config file support previously - is the issue just that it didn't fail previously?

From the original issue it seems like the actual problem was that gomplate was running from a directory it had no read permission for?

If that's the case, the error makes sense, but I'm not sure about the fix. What if the error is legitimate? i.e. what if the user has a config present but are running as the wrong user with no read permissions, or is in the incorrect directory?

I suppose what we need here is some way to signal to gomplate that it shouldn't attempt to read a config file at all. Perhaps setting `--config=""` to override the default of `.gomplate.yaml`?

nabokihms:
> is the issue just that it didn't fail previously?
Issue appeared straight after upgrade.

> From the original issue it seems like the actual problem was that gomplate was running from a directory it had no read permission for?

Yeah, I managed to figure it out after reading the code. My home dir was the root of a distroless image, which was not accessible.

I'm totally ok with the empty config flag, for me it solves the problem.

hairyhenderson:
I've merged https://github.com/hairyhenderson/gomplate/pull/2227 just now, which should solve the underlying issue here. Closing now.

Issue #2182, opened 2024-07-18
Is the second example for `coll.Merge` incorrect?


cofiem:
The [second example for `coll.Merge`](https://docs.gomplate.ca/functions/coll/#collmerge):

```
$ gomplate -i '{{ $dst := dict "foo" 1 "bar" 2 }}
{{ $src1 := dict "foo" 8 "baz" 4 }}
{{ $src2 := dict "foo" 3 "bar" 5 }}
{{ coll.Merge $dst $src1 $src2 }}'
map[foo:1 bar:5 baz:4]
```

https://github.com/hairyhenderson/gomplate/blob/19cd050be438f9d407800725f16139d8a5f86624/docs-src/content/functions/coll.yml#L403-L407

https://github.com/hairyhenderson/gomplate/blob/19cd050be438f9d407800725f16139d8a5f86624/docs/content/functions/coll.md?plain=1#L621-L625

But using gomplate `v4.1.0` I get:

```
$ gomplate -i '{{ $dst := dict "foo" 1 "bar" 2 }}
> {{ $src1 := dict "foo" 8 "baz" 4 }}
> {{ $src2 := dict "foo" 3 "bar" 5 }}
> {{ coll.Merge $dst $src1 $src2 }}'

map[bar:2 baz:4 foo:1]
```

The value for `bar` is `2`?

hairyhenderson:
Looks like #2217 will fix this

PR #2217, opened 2024-08-30
docs(fix): Fix coll.Merge example

live851738909:
It should be written wrong here, the priority order is from left to right, the value of bar should be 2

Fies #2182


hairyhenderson (approved):
LGTM - thanks

hairyhenderson:
You'll need to change the commit message to something like `docs(fix): Fix coll.Merge example`, and fix the source file for the document first.

PR #2238, opened 2024-10-09
Minor fix in documentation related to `--exclude-processing`

assayire:
Using backticks in the gomplate like so will result in error: `*.png: command not found`:
```
$ gomplate --exclude-processing `*.png` --input-dir in/ --output-dir out/
```

Fixed the command:

```
$ gomplate --exclude-processing '*.png' --input-dir in/ --output-dir out/
```


hairyhenderson (approved):
Thanks!

PR #2240, opened 2024-10-20
feat: Added ssh namespace

libmonsoon-dev:
Tested on Linux. I would be grateful if someone could see if it works on Windows, I don’t have that option

### DIFF
--- a/docs-src/content/functions/crypto.yml
+++ b/docs-src/content/functions/crypto.yml
@@ -9,6 +9,47 @@ preamble: |
   recommended to have your resident security experts inspect gomplate's code
   before using gomplate for critical security infrastructure!_
 funcs:
+  - name: crypto.SSH
+    released: v4.2.0 # I hope so
+    description: |
+      Namespace for ssh functions
+    pipeline: false
+    examples:
+      - |
+        $ gomplate -i '{{ crypto.ssh }}'
+        <namespace SSH [PublicKey]>
+
+  - name: crypto.SSH.PublicKey
+    released: v4.2.0
+    description: |
+      Loads [Secure Shell](https://en.wikipedia.org/wiki/Secure_Shell) public key
+    pipeline: true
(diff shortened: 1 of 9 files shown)


hairyhenderson:
Hi @libmonsoon-dev, sorry for the delay in reviewing.

For major features like this I prefer if you file an issue first so we can discuss the specifics before you spend time on it. See https://github.com/hairyhenderson/gomplate/blob/main/CONTRIBUTING.md for details.

It seems to me that you've added a new SSH sub-namespace inside the `crypto` namespace? Is there a reason for that? It's unusual and has no precedent.

From a quick scan at what you've added, it seems most of these should just be functions in the `crypto` namespace, since all they're doing is reformatting public keys into OpenSSH-compatible formats.

Can you explain in greater detail what you need to accomplish?

hairyhenderson:
> Okay, next time I'll create an issue first. But is it necessary to create one in this case, or can we discuss these changes right here?

We can discuss here - just keep in mind that I may require significant changes for this to be merged if it's merged at all (no guarantees).

> Yes, that's right. I wanted to recreate the structure similar to the packages in the standard library (and in golang.org/x/...). There ssh is separated into a subpackage, and in my opinion that makes sense.

gomplate's function namespaces are not really intended to mirror the stdlib. There's only a single level of namespaces currently, and I'm not convinced there's a compelling reason to support additional layers.

> I would like the template not to contain a specific key, but instead, when generating the config from the template, the current key of the user who started the generation will be used

The goal is still a bit unclear to me, but it seems to me that you probably don't need to parse the public key at all, but rather to simply [`include`](https://docs.gomplate.ca/functions/data/#include) it?

Wouldn't something like `gomplate -d mykey=$HOME/.ssh/id_rsa.pub -I '{{ include "mykey" | base64.Encode }}'` work for this use-case?

PR #2283, opened 2024-12-14
ci(bench): Add benchmark-action

hairyhenderson:


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Run benchmark
+        run: make bench.txt
+      - name: Download previous benchmark data
+        uses: actions/cache@v4
+        with:
+          path: ./cache
+          key: ${{ runner.os }}-benchmark
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
```
thaJeztah, line 26: TIL about this action. Sounds interesting!

(sorry for the random comment, I'm still subscribed to this repo, just a bit of a fly on the wall here ❤️)

hairyhenderson: 👋😁

I've been toying around with it a bit in a few repos... It can generate some graphs (like https://hairyhenderson.github.io/go-codeowners/dev/branch), but I don't like how they look. I've been thinking of pushing the data into a public Grafana dashboard; maybe a Christmas vacation side-project 😂

thaJeztah: Same(ish); we've been discussing for a while that we would really love to have metrics of CI over time; (performance) regressions always tend to sneak in really granularly, and honestly, GitHub actions does not provide any good tooling on itself to keep eyes on those.

The BuildKit team started some work on a "bench" repository for BuildKit to compare things between versions;

- https://github.com/moby/buildkit-bench
- website with results; https://moby.github.io/buildkit-bench/

(for sure "borrow" anything from that if its of use! we may be doing so as well for the engine 😄)

thaJeztah: (We also have some internal tools that collect data, which I think is also a combination of grafana and other bits and bops, but not yet in a good shape for this exact purpose)

Issue #2301, opened 2025-01-15
Adding output files deletion functionality


unlimitedbits:
Hi,

we are considering using gomplate for rendering configurations for infrastructure rollouts (e.g. outputting terraform files).
As we want to support the full lifecycle of such components, it would also be handy if gomplate was able to just delete the output files of a templating step, especially as we are using outputMap and the actual output path for the input templates is different based on the datasources and the target folder potentially also consists of other files.

From our side two options could really help here:
* Adding a --delete option to the CLI so that gomplate is deleting the output files instead of rendering
* Adding something like a --dry-run option which prints the paths of all files which would be rendered

Would you consider adding one of those options to gomplate?

hairyhenderson:
Can you elaborate on the `--delete` option? I'm not sure I follow the logic.

As for `--dry-run`, there are many ways that gomplate can output files, and fully rendering the output is really the only correct way to know what'll happen.

Generally for this sort of thing I would recommend tracking these rendered files with a version control system (e.g. `git`), and then you can look at the state of the workspace post-render (i.e. `git diff`) to see what's changed.

mihkels:
Sorry for the delayed response. Consider a scenario where we need to generate multiple outputs from the same source templates based on a dynamically configured `outputMaps`. The challenge arises because the output directories are determined dynamically, making it difficult to track the generated files without parsing the `outputMaps`. However, the values within `outputMaps` are not accessible during the template generation process or after processing.

Example:

```go
  {{ root_dir_from_dynamic_config }}/app/{{ configurable }}/platform/app-name/{{ .in | strings.ReplaceAll ".yaml.tmpl" ".yaml" }}
```

In this example, `root_dir_from_dynamic_config` and `configurable` are dynamically configured values from `datasources`. There can be more than two dynamically configurable values. Currently, there is no way to access the generated output paths.

hairyhenderson:
@mihkels thanks for the response. If I'm understanding correctly, being able to access the output file paths from within the template's data might help. That way you could generate a manifest of sorts to track the files that were rendered.

However, I'm still not sure why using git and watching which files were modified wouldn't help...

unlimitedbits:
@hairyhenderson thanks for your inputs

[...]

PR #2302, opened 2025-01-16
go(feat): Add the dry-run flag to gomplate (#2301)

mihkels:


hairyhenderson:
@mihkels Thanks for working on this, can you give more details on what the goal is here?

In general I'd prefer an issue to be filed first so we can discuss before any time is spent on writing code.

hairyhenderson:
..a month later, sorry for the oversight, I see that https://github.com/hairyhenderson/gomplate/issues/2301 was indeed filed first. Given there was no discussion on that issue, this PR may have been premature. When I have some time I'll review #2301 to see if that's actually functionality that makes sense.

PR #2429, opened 2025-08-22
feat(crypto): New experimental crypto.PBKDF2MCF function

vic1707:
Added MCF version for pbkdf2 

Maybe this version could use some default parameters and a random salt since these are returned in the MCF format 🤔


hairyhenderson:
@vic1707 thanks for the contribution - can you fix the PR title (and main commit message) to something like `feat(crypto): New crypto.PBKDF2MCF function`?

PR #2439, opened 2025-09-19
Replace AWS SDK v1 with v2

sagikazarmark:
The v1 SDK has been deprecated as of July: https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

I also removed the EC2 metadata region detection: it's implemented in SDK v2.


hairyhenderson:
the benchmark's not a concern - it's not properly set up for PRs from forks...

One concern I have is that this changes exported API, so is technically a breaking change and should trigger a v5 release. I'm digging into this a bit more to see if we can avoid that.

PR #2550, opened 2026-03-25
ci(docker): speed up broken/slow docker build

hairyhenderson:


thaJeztah:
👋 let me know if you need any help (I could look for some folks); if you're looking to simplify, there's now also a new combined action / reusable flow that was created recently; https://docs.docker.com/build/ci/github-actions/github-builder/

hairyhenderson:
Ok, finally found time to get it working with the `docker/github-builder` action

PR #2553, opened 2026-03-31
fix(conv): Print actual argument type in InterfaceSlice error

ajpetersons:
This fixes an issue where `conv.Slice` would produce a misleading error message if the argument could not be parsed as a slice. Previously the error message would indicate that the type provided was `reflect.Value` which was not necessarily the case.


hairyhenderson (approved):
Thanks for the fix, @ajpetersons!

PR #2610, opened 2026-05-02
ci(release): replace deploy-docker with docker/github-builder reusable workflow

hairyhenderson:
## Summary

- Replaces the bespoke `deploy-docker` job in `release.yml` with the same `docker/github-builder` reusable workflow used in `docker-build.yml`
- A `deploy-docker-version` pre-job strips the `v` prefix and extracts the major version string, then `deploy-docker-latest` and `deploy-docker-alpine` run in parallel
- `VERSION` is passed explicitly as a build arg so the binary version is correct regardless of git state inside the Docker build
- Tags pushed per release to all three registries (`ghcr.io/hairyhenderson/gomplate`, `gomplate/gomplate`, `hairyhenderson/gomplate`): `latest`, `stable`, `v5.1.0`, `v5` (and `-alpine` variants)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/create-github-app-token@v3
         id: app-token
         with:
-          app-id: ${{ vars.APP_ID }}
+          client-id: ${{ vars.APP_ID }}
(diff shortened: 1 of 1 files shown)


thaJeztah:
you got it working? nice!

PR #2791, opened 2026-08-24
feat(net): add net.GenerateMAC function

### LINKED ISSUE #1699: Function to generate MAC addresses
Please add a function that allows generating MAC addresses based on some value (similar to [ansible's community.general.random_mac](https://docs.ansible.com/ansible/devel/collections/community/general/random_mac_filter.html) when used with a seed)

ChrisJr404:
This adds the `net.GenerateMAC` function from #1699, following the shape you sketched out in that thread.

With no arguments it returns a random, locally administered unicast address (local bit set, multicast bit cleared) so it stays out of real vendor ranges. A single argument fixes the leading octets as a prefix, and a second argument seeds the result so the same prefix and seed always produce the same address. Separators in the prefix are ignored, so `aa:bb`, `aa-bb`, and `aabb` are all treated the same.

Tests and docs are included. I left it out of the experimental gate since the surface is small and matches what was discussed, but happy to put it behind that if you would prefer to iterate on it first.

Fixes #1699

### DIFF
--- a/docs-src/content/functions/net.yml
+++ b/docs-src/content/functions/net.yml
@@ -131,6 +131,37 @@ funcs:
         [
           "v=spf1 -all"
         ]
+  - name: net.GenerateMAC
+    released: v5.3.0
+    description: |
+      Generate a MAC (hardware) address, returned as a string in the usual
+      colon-separated form (e.g. `aa:bb:cc:dd:ee:ff`).
+
+      With no arguments a random locally administered unicast address is
+      generated, so it won't collide with a real vendor-assigned address.
+
+      An optional `prefix` fixes the leading octets of the address (such as an
+      OUI); the rest are filled in randomly. Colons, hyphens, and dots in the
+      prefix are ignored, so `aa:bb`, `aa-bb`, and `aabb` are all equivalent.
+
+      An optional `seed` makes the output deterministic: the same prefix and
+      seed always produce the same address. Use an empty prefix to seed the
+      whole address.
+    pipeline: true
+    arguments:
+      - name: prefix
+        required: false
+        description: The fixed leading octets of the address. Defaults to empty (a fully-generated address).
+      - name: seed
+        required: false
+        description: When provided, the address is derived deterministically from this value instead of being random.
+    examples:
+      - |
+        $ gomplate -i '{{ net.GenerateMAC }}'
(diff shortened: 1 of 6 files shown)


hairyhenderson:
Thanks @ChrisJr404, this looks great!

Just one thought - while MAC addresses with `:` separators and lower-case letters are quite common, the notation standard is actually `-`-separators and upper-case letters:

> The standard ([IEEE 802](https://en.wikipedia.org/wiki/IEEE_802)) format for printing EUI-48 addresses in human-friendly form is six groups of two [hexadecimal](https://en.wikipedia.org/wiki/Hexadecimal) digits, separated by hyphens (-) in transmission order (e.g. 01-23-45-67-89-AB). This form is also commonly used for EUI-64 (e.g. 01-23-45-67-89-AB-CD-EF).[[3]](https://en.wikipedia.org/wiki/MAC_address#cite_note-eui_tutorial-3) Other conventions include six groups of two hexadecimal digits separated by colons (:) (e.g. 01:23:45:67:89:AB), and three groups of four hexadecimal digits separated by dots (.) (e.g. 0123.4567.89AB); again in transmission order.

_From https://en.wikipedia.org/wiki/MAC_address_

I wonder if it'd make sense to allow altering the output form in the function?

One thought I had was using the prefix as a hint (i.e. if the prefix is "01-23-FF", use uppercase and hyphens), but it seems like it won't work well (what if the user doesn't want a prefix? what if the prefix uses only digits 0-9?).

Do you have any thoughts?

ChrisJr404:
Good point on the notation. My inclination is to keep the default output as lowercase colon-separated, since that is what most tooling and configs expect in practice, and not try to infer the form from the prefix, for exactly the reasons you gave (an all-digit or absent prefix gives no signal, and guessing would surprise people).

If altering the form is worth supporting, I would rather make it explicit than inferred, e.g. an optional style argument like `net.GenerateMAC "hyphen"` / `"dot"` defaulting to colon. Happy to add that here if you want it in this PR, or keep this one focused on generation and leave formatting to a follow-up. Which do you prefer?

hairyhenderson:
I'd be happy with a followup to add formatting. A style argument sounds good enough!

hairyhenderson (approved):
LGTM!

PR #2792, opened 2026-08-24
feat(net): add net.GenerateMAC for generating MAC addresses

ChrisJr404:
This adds net.GenerateMAC, following the shape you sketched out in #1699. With no arguments you get a random, locally administered unicast address, a single argument fixes a prefix, and a second argument acts as a seed so the same prefix and seed always produce the same address.

Docs and tests are included. The prefix accepts `:` or `-` separators.


ChrisJr404:
Closing this one, I realized I already have #2791 open for the same function. Sorry for the double, let's keep the discussion on #2791.

